Issue #86, opened 2018-02-05
Create igxLabel, igxInput user story and update current topics based on the article guidance


bkulov:
Combine igxLabel and igxInput into one topic.

Article guidance - https://github.com/IgniteUI/igniteui-docfx/blob/master/.github/CONTRIBUTING.md

bkulov:
Closing this issue as it's a duplicate of #87.

sstoyanovIG:
All done

PR #87, opened 2018-02-05
Create Label and Input Demo.

sstoyanovIG:
Create Label and Input Demo. 

#86


kdinev:
@zdrawku Liz will work on translations next week and she will pull directly to master.

PR #74, opened 2018-01-26
Reworked angular slider topic

sboykova:
Issue #46

PR #89, opened 2018-02-06
Translated rewritten slider topic

ig-ealbert:
Translated the raw file as provided in the slider-docs-rework branch.

One thing I noticed is that the language metadata is removed.  Is that intentional?  It's removed in the rewritten English topic and I simply copied what was there.

### DIFF
--- a/jp/components/slider.md
+++ b/jp/components/slider.md
@@ -5,56 +5,215 @@ _keywords: Ignite UI for Angular, UI コントロール, Angular ウィジェッ
 ---
 
 ## Slider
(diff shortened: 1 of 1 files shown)


kdinev:
@simeonoff Can you answer Liz' question?

ig-ealbert:
Now that #74 has been merged, what do I do with this pull request?

kdinev:
@ig-ealbert I can merge it. It won't make a difference.

ig-ealbert:
Redid this in #97.

PR #96, opened 2018-02-08
add topic for toggle

### LINKED ISSUE #95: Create topic for IgxToggle directive
Create topic for Toggle Directive

Aleksandyr:
Closes #95


ig-ealbert:
@zdrawku I will work on this today and will add the topic once it is reviewed.

ig-ealbert:
I added the Japanese topic.  Please note that the demos are just placeholders, so if you're planning to publish soon, those should be filled in with the real demos.
Also, "So let's deep in" does not make sense.  Possible suggestions:
1. "So let's dive in"
2. "So let's begin"
3. "So let's jump in"
4. "So let's get started"

Aleksandyr:
Thanks for the feedback. I've made the change you suggested. As for the demos they are in progress. When they are implemented we will add them into the topic.

PR #115, opened 2018-02-16
Add igxForOf topic

dkamburov:
Related https://github.com/IgniteUI/igniteui-angular/issues/552

### DIFF
--- a/en/components/for_of.md
+++ b/en/components/for_of.md
@@ -108,6 +108,12 @@ The directive can be used to virtualize the data in vertical, horizontal or both
 </div>
 ```
 
+**Note**: The `igxForOf` directive can bind to remote service too. Review the following demo for more information.
+#### igxFor bound to remote service Demo
+<div class="sample-container loading" style="height:780px">
+    <iframe src='{environment:demosBaseUrl}/igx-for-sample-2' width="100%" height="100%" seamless frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
+</div>
+<br/>
 <div class="divider--half"></div>
 
 ## API
--- a/en/components/for_of.md
+++ b/en/components/for_of.md
@@ -130,11 +130,11 @@ Below is the list of all inputs that the developers may set to configure the `ig
 
 A list of the events emitted by the **igx-for**:
 
-| Name              | Description                                                              |
-| :---------------- | :----------------------------------------------------------------------- |
-| _Event emitters_  | _Notify for a change_                                                    |
-| `onItemLoad`      | Used when scrolled to emit the loaded chunk of data                      |
-| `onItemPreload`   | Used when scrolled to emit the chunk of data which is about to be loaded |
+| Name              | Description                                                          |
+| :---------------- | :------------------------------------------------------------------- |
+| _Event emitters_  | _Notify for a change_                                                |
+| `onItemLoad`      | Used when scrolled to emit the loaded data item                      |
+| `onItemPreload`   | Used when scrolled to emit the data item which is about to be loaded |
 
 <div class="divider"></div>
 
--- a/en/components/for_of.md
+++ b/en/components/for_of.md
@@ -8,14 +8,15 @@ _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI 
 
 <p class="highlight">
 `igxForOf` is now available as an alternative to `ngForOf` for templating large amounts of data. The `igxForOf` uses virtualization technology behind the scenes to optimize DOM rendering and memory consumption. Virtualization technology works similar to Paging by slicing the data into smaller chucks which are swapped from a container viewport while the user scrolls the data horizontally/vertically. The difference with the Paging is that virtualization mimics the natural behavior of the scrollbar.
-The directive is creating scrollable containers and renders small chunks of the data. Is is used inside igx-grid and it can be used to build virtual igx-list.
+The directive is creating scrollable containers and renders small chunks of the data. It is used inside `igx-grid` and it can be used to build virtual `igx-list`.
+`igxForOf` can be used to virtualize data horizontally, vertically or both.
 </p>
 <div class="divider"></div>
 
 ### igxFor Demo
 
 <div class="sample-container loading" style="height:780px">
-    <iframe src='{environment:demosBaseUrl}/igx-for' width="100%" height="100%" seamless frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
+    <iframe src='{environment:demosBaseUrl}/igx-for-sample-1' width="100%" height="100%" seamless frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
 </div>
 <br/>
 <!--<button data-localize="stackblitz" class="stackblitz-btn">view on stackblitz</button> -->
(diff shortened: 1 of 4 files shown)


dkamburov:
@mpavlinov I'm good with them, the overview section is great.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+        ...
+        IgxForOfModule,
+        ...
+    ]
+})
+export class AppModule {}
+```
+
+### Usage
+
```
mpavlinov: @dkamburov We should add 3 sections here for horizontal, vertical and both virtualization.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+### Inputs
+
+Below is the list of all inputs that the developers may set to configure the `igxFor` look/behavior:
+| Name | Type | Description |
+| :--- |:--- | :--- |
+| id | string | Unique identifier of the Grid |
+| `igxForOf` | any[] | The data to be virtualized |
+| `igxForScrollOrientation` | string | Specifies the scroll direction - "horizontal" or "vertical", default value is _vertical_ |
+| `igxForScrollContainer` | any | Specifies the container where the helper scrollbars will be contained, it is useful when nesting the directive |
```
mpavlinov: @dkamburov The description is very generic. I don't understand how to use this option. We should make the description better. Let's discuss this tomorrow.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+| `igxForContainerSize` | any |  Specifies the container size in pixels. For "vertical" virtualization this is the height of the container. For "horizontal" virtualization this is the width of the container. |
+| `igxForItemSize` | any | Specifies the item size in pixels. For "vertical" virtualization this is the height of the item/row.  |
```
mpavlinov: @dkamburov Is it only applicable for vertical virtualization? If so, we should explicitly state it in the docs.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+
+<div class="divider"></div>
+
+### Methods
+
+Here is a list of all public methods exposed by the **igx-for**:
+
+| Signature       | Description                     |
+| :-------------- | :------------------------------ |
+| `scrollNext()`  | Loads the previous chunk of data |
```
mpavlinov: @dkamburov Maybe "nextChunk"/"previousChunk" names are more appropriate. I'll propose it as a change in the API.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+<div class="divider--half"></div>
+
+### Outputs
+
+A list of the events emitted by the **igx-for**:
+
+| Name              | Description                                                              |
+| :---------------- | :----------------------------------------------------------------------- |
+| _Event emitters_  | _Notify for a change_                                                    |
+| `onChunkLoaded`   | Used when scrolled to emit the loaded chunk of data                      |
```
mpavlinov: @kdinev From what I know we're not using "ed"/"ing" events anymore. We have only one event to keep the event number lower across the product. Should we remove one of the events?

kdinev: @mpavlinov Correct. @rkaraivanov should chip in on what the correct naming should be. I guess `onChunkLoad`

en/components/for_of.md
kdinev: Make the wording such that it's clear what it means in horizontal and vertical scenarios.

en/components/for_of.md
MayaKirova: @kdinev @mpavlinov We're actually using both events at the moment in the grid. The "ing" event is used to clean some column/row states in certain scenarios, while the "ed" event is used for the keyboard navigation implementation.
From the directive's perspective itself we definitely need an event before the new chunk is actually loaded to provide  the users with the capability implement remote virtualization. If it's absolutely neccessary to move away from "ing"/"ed" naming convention then we could name one onScroll and the other onChunkLoad. Let me know what you think.

en/components/for_of.md
```diff
@@ -0,0 +1,97 @@
...
+## API
+
+### Inputs
+
+Below is the list of all inputs that the developers may set to configure the `igxFor` look/behavior:
+| Name | Type | Description |
+| :--- |:--- | :--- |
+| id | string | Unique identifier of the Grid |
+| `igxForOf` | any[] | The data to be virtualized |
+| `igxForScrollOrientation` | string | Specifies the scroll direction - "horizontal" or "vertical", default value is _vertical_ |
```
MayaKirova: We currently don't set a default value for igxForScrollOrientation.

en/components/for_of.md
dkamburov: @mpavlinov Actually the behavior is not described correctly, the methods are scrolling by one item in the appropriate direction, so it's probably better to leave the names as they are and change the description

en/components/for_of.md
```diff
@@ -0,0 +1,149 @@
+---
+title: igxFor
+_description: The Ignite UI for Angular product is providing virtual directive igxFor similar to ngFor, but it is creating scrollbars and only renders small chunks of the data.
```
kdinev: Ignite UI for Angular now exposes a virtual igxFor directive similar to ngFor, which virtualizes DOM object rendering by visualizing only the visible chunks of the data in the DOM.

en/components/for_of.md
```diff
@@ -0,0 +1,149 @@
...
+_keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI widgets, Angular, Native Angular Components Suite, Native Angular Controls, Native Angular Components Library, Virtualization, Performance, Virtual directive, Virtual For
+---
+
+## igxFor
+
+<p class="highlight">Display huge amount of repeated items. The directive is creating scrollable containers and renders small chunks of the data. Is is used inside igx-grid and it can be used to build virtual igx-list.</p>
```
kdinev: Virtualizes DOM object rendering by visualizing only the visible chunks of the data in the DOM. DOM objects are reused while the end-user is scrolling. The igx-grid implements the igxFor directive internally, but the directive can also be implemented on any component that needs to render large amounts of data, like the igx-list.

en/components/for_of.md
```diff
@@ -0,0 +1,149 @@
...
+
+```html
+<span #container>
+    <ng-template *igxFor="data"></ng-template>
+</span>
+```
+
+
+
+The **data** property is an array and it provides the virtualized data.
```
kdinev: The date itself is not virtualized, so just call it "the data"

en/components/for_of.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+The directive can be used to virtualize the data in vertical, horizontal or both directions.
```
kdinev: virtualized the DOM?

en/components/for_of.md
kdinev: `onChunkPreload` and `onChunkLoad`?

en/components/for_of.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Outputs
+
+A list of the events emitted by the **igx-for**:
+
+| Name              | Description                                                          |
+| :---------------- | :------------------------------------------------------------------- |
+| _Event emitters_  | _Notify for a change_                                                |
+| `onItemLoad`      | Used when scrolled to emit the loaded data item                      |
+| `onItemPreload`   | Used when scrolled to emit the data item which is about to be loaded |
```
kdinev: I'm waiting on agreement on the PR to the main repository to approve the final API.
https://github.com/IgniteUI/igniteui-angular/pull/619

PR #120, opened 2018-02-19
Eaa rewritten topics

ig-ealbert:
Added reviewed topics from various rewrites.  These were based on pull requests that were assigned to me.

Additionally, I went through the history of the EN folder to try to get anything that was done without any assignment for localization (and therefore no notification).  This includes some updates for StackBlitz demos.

### DIFF
--- a/jp/components/button.md
+++ b/jp/components/button.md
@@ -12,25 +12,119 @@ _language: ja
 
 ### Button デモ
 
-`igxButton` を使用して span 要素を Ignite UI for Angular のスタイル設定済みボタンにします。
-
-```html
-<span igxButton="raised" igxButtonColor="yellow" igxButtonBackground="#000">クリック<span>
-```
-
-<div class="sample-container loading" style="height: 1192px">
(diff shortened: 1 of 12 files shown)


jp/components/button.md
```diff
@@ -12,26 +12,120 @@ _language: ja
...
+
+<div class="sample-container" style="height: 48px">
+    <iframe seamless width="100%" height="100%" frameborder="0" src="{environment:demosBaseUrl}/buttons-sample-3">
+</div>
+
 `igxButton` を使用して span 要素を Ignite UI for Angular のスタイル設定済みボタンにします。
 
 ```html
-<span igxButton="raised" igxButtonColor="yellow" igxButtonBackground="#000">クリック<span>
+<span igxButton="raised" igxButtonColor="yellow" igxButtonBackground="#000" igxRipple="yellow">Click me</span>
```
kdinev: Shouldn't this one be in Japanese as well?

ig-ealbert:
@kdinev You are correct - it should be consistent.  In other topics I stopped translating the text used in the code snippets because then they wouldn't match the samples, which are in English.  When the samples are localizable I may revisit this, but for now, as long as it's consistent within the topic it's ok.  Thanks for catching that.

PR #140, opened 2018-02-22
Refactor dialog topic and add api summary

### LINKED ISSUE #124: Create Dialog Windows user story and update current topic based on the article guidance.
Article guidance - https://github.com/IgniteUI/igniteui-docfx/blob/master/.github/CONTRIBUTING.md

sstoyanovIG:
Closes #124

### DIFF
--- a/en/components/dialog.md
+++ b/en/components/dialog.md
@@ -4,12 +4,12 @@ _description: Whether building informative dialogs or data manipulation windows,
 _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI widgets, Angular, Native Angular Components Suite, Native Angular Controls, Native Angular Components Library, Angular Dialog Window components, Angular Dialog Window controls
 ---
 
-##Dialog Windows
+##Dialog Window
 <p class="highlight">Use the Ignite UI for Angular Dialog Window component to display messages or present forms for users to fill out. The component opens a dialog window centered on top of app content. You can also provide a standard alert message that users can cancel.</p>
 <div class="divider"></div>
 
-### Dialog Example
-<div class="sample-container loading" style="height:588px">
+### Dialog Demo
+<div class="sample-container loading" style="height:300px">
     <iframe id="dialog-sample-iframe" src='{environment:demosBaseUrl}/dialog' width="100%" height="100%" seamless frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
 </div>
 <div>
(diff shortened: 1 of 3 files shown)


kdinev:
@ig-ealbert Stefan has overwritten some of the JP content, where it gets updated. Can you double check though?

en/components/dialog.md
```diff
@@ -4,87 +4,180 @@ _description: Whether building informative dialogs or data manipulation windows,
...
+</div>
+<div class="divider--half"></div>
 
-`leftButtonBackgroundColor`
+####Custom Dialog
 
-`leftButtonRipple`
-<div class="divider--half"></div>
+To add custom dialog, in the template of our sign in component we can add the following code to get the custom dialog. We have to set the `title`,`leftButtonLabel`,`rightButtonBackgroundColor`, `rightButtonColor`, `rightButtonLabel`, `closeOnOutsideSelect` and handle `onLeftButtonSelect` and `onRightButtonSelect` event.
+Also we can add two groups of label and input decorated with the [**igxLabel**](../components/label_input.html) and [**igxInput**](../components/label_input.html) directives.
```
ig-ealbert: I've now seen links done in 3 ways in the topics:
1. Production link.
2. Relative link that is just the name of the topic.
3. Relative link as is used here.
Which is correct?  Do all of them work?  Should it be consistent?

en/components/dialog.md
```diff
@@ -4,87 +4,180 @@ _description: Whether building informative dialogs or data manipulation windows,
...
+#### Inputs
+The following inputs are available in the **igx-dialog** component:
+
+| Name | Type | Description |
+| :--- | :--- | :--- |
+| `id` | string | Unique identifier of the dialog. |
+| `title` | string | Specifies the title of the dialog. |
+| `message` | string | Specifies the message of the dialog. |
+| `leftButtonLabel` | string | When set, the left button becomes visible. |
+| `leftButtonType` | string | Specifies the left button [**type**](button.md#button-types). Default value is flat. |
```
ig-ealbert: I'm pretty sure you can't link to .md files in production.

jp/components/dialog.md
```diff
@@ -1,99 +1,183 @@
 ﻿---
-title: Dialog Window コンポーネント
-_description: Ignite UI for Angular Dialog Window コンポーネントを使用すると、情報ダイアログまたはデータ変換ウィンドウを作成し、リアルタイムで情報を表示して管理できます。
-_keywords: Ignite UI for Angular, UI コントロール, Angular ウィジェット, web ウィジェット, UI ウィジェット, Angular, ネイティブ Angular コンポーネント スィート, ネイティブ Angular コントロール, ネイティブ Angular コンポーネント ライブラリ, Angular Dialog Window コンポーネント, Angular Dialog Window コントロール
-_language: ja
```
ig-ealbert: Did we stop using this metadata field?

en/components/dialog.md
sstoyanovIG: The correct way is to use .md files. I will update these ones.

en/components/dialog.md
sstoyanovIG: Basically there is a parser, which converts the .md files to html files, so this should work on production

jp/components/dialog.md
sstoyanovIG: I noticed none of the topics has this field, may be @simeonoff know more?

simeonoff: These are required fields, each article should include them. The values for _description and _keywords are provided by marketing. Simply put - you shouldn't be deleting them.

en/components/dialog.md
zdrawku: We are going to review the changes after merging them to staging.

jp/components/dialog.md
ig-ealbert: @simeonoff I meant particularly the `_language` metadata field.  That is the one that is being removed in some of the new topics.  I think it's a copy/paste error as I've noticed the developer copying the English topic into the Japanese file.  I have been fixing this as I add the Japanese translations but it's error prone as I have to remember to do it.

ig-ealbert: Not sure if it needs to be in there for the English topic.  Mostly interested if it is still used in the Japanese topics to know if I should continue adding it in there.

Issue #121, opened 2018-02-20
igxTabBar demo refactoring


gedinakova:
Update igxTabBar demo to:
 - Include a step by step guide to using the component, and
 - Reflect api changes(if any).

PR #143, opened 2018-02-22
Updating the TabBar demo page

rmkrmk:
#121


ig-ealbert:
I can't find any commits by @rmkrmk in the list of commits.  What should I review?  This just looks like a bunch of merges.

PR #190, opened 2018-03-02
Mostly link fixes

ig-ealbert:
1. Changed internal anchors to use Japanese text where applicable (so they work).
2. Changed production URLs to be relative (so they work on staging too).
3. Added rewritten NavBar topic.

### DIFF
--- a/jp/components/list.md
+++ b/jp/components/list.md
@@ -81,10 +81,10 @@ export class AppModule {}
 <!--contacts.component.html-->
 
 <igx-list>
-    <igx-list-item isHeader="true">ヘッダー</igx-list-item>
-    <igx-list-item>項目 1</igx-list-item>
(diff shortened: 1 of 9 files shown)


jp/components/list.md
```diff
@@ -81,10 +81,10 @@ export class AppModule {}
 <!--contacts.component.html-->
 
 <igx-list>
-    <igx-list-item isHeader="true">ヘッダー</igx-list-item>
-    <igx-list-item>項目 1</igx-list-item>
-    <igx-list-item>項目 2</igx-list-item>
-    <igx-list-item>項目 3</igx-list-item>
+    <igx-list-item isHeader="true">Header</igx-list-item>
```
ig-ealbert: In case there is any question, I did this on purpose so that it would match the sample displayed below it.  The samples are not internationalized so they will all appear in English.

PR #214, opened 2018-03-15
Adding Column Pinning to igxGrid doc

ChronosSF:


en/components/grid.md
```diff
@@ -544,6 +544,40 @@ public ngOnInit() {
 }
 ```
 
+### Column Pinning
+
+**Column Pinning** is available through the **igx-grid** API. Each column can be pinned as long as the pinned area does not become wider than the grid itself. Column pinning is controlled through the `pinned` input of the `igx-column`. Pinned columns are always rendered on the left side of the grid and stay fixed through horizontal scrolling of the unpinned columns in the grid body.
```
kdinev: Please provide a concrete example with dimensions of when the pinning would be refused.

PR #229, opened 2018-03-26
fix dialog snippet with correctly bound property

Aleksandyr:


zdrawku:
Related to https://github.com/IgniteUI/igniteui-angular-samples/issues/132

Issue #230, opened 2018-03-28
Japanese translation suggestion


DaizenIG:
# Description

I would like to suggest the translation for JP to make it better meaning to developers.

# Step to reproduce
- Go to EN Grid documentation: https://github.com/IgniteUI/igniteui-docfx/blob/master/en/components/grid.md
In Filtering, Note says the following:
[!NOTE] The filtering operation DOES NOT change the underlying data source of the grid.

- Go to JP Grid documentation: https://github.com/IgniteUI/igniteui-docfx/blob/master/jp/components/grid.md
In Filtering, Note says the following:
[!NOTE] フィルタリング操作でグリッドの基になるデータ ソースは変更しません。

# Expected Result

I'm suggesting the translation of the sentence below:　

[!NOTE] フィルタリング操作では、グリッドにバインドされているデータ ソースは変更されません。

I see the same expressions for following features
- Sorting

ig-ealbert:
This should be fixed with #232.

PR #231, opened 2018-03-28
Update grid.md

yoshimis:
#230 Updated JP according to the suggestion.


ig-ealbert:
Yoshimi says that this pull request is incorrect and that #232 should be used instead.

Issue #228, opened 2018-03-26
Break down igxGrid doc into separate docs per feature


zdrawku:
Each grid feature should be in separate doc file

PR #243, opened 2018-04-04
Breakdown main grid topic to separate smaller topics

### LINKED ISSUE #233: Create a separate igxGrid virtualization topic
Related to #228 

The topic should include the following: 
- Explanation when the grid will be virtualized
- Limitations to such grids
- What is required for virtualization, why width expected and enforced, why can't columns grow through templating outside the set width. 
- Examples for combinations of set width/heights along set widths of columns heights for rows
- Stackblitz link for the performance sample

zdrawku:
Closes #233 #228 

@ig-ealbert I have edited the JP files as well, although they may need some further modifications.


jp/components/grid_column_pinning.md
```diff
@@ -0,0 +1,56 @@
...
+
+```typescript
+public columnPinning(event) {
+    if (event.column.field === "Name") {
+        event.insertAtIndex = 0;
+    }
+}
+```
+
+<div class="divider--half"></div>
```
ig-ealbert: This line doesn't exist in the English version of the pinning topic.

jp/components/grid_column_pinning.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Grid column pinning
+_description: The Ignite UI for Angular Data Grid control features the fastest, touch-responsive data-rich grid with popular features, including hierarchical and list views.
+_keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI widgets, Angular, Native Angular Components Suite, Native Angular Controls, Native Angular Components Library, Angular Data Grid component, Angular Data Grid control, Angular Grid component, Angular Grid control, Angular High Performance Grid, column pinning, pinning, pin
+---
```
ig-ealbert: I think that the Japanese topics are supposed to have a `_language: ja` entry in the metadata?  This would apply to all of the new topics here.

zdrawku: Yes, I just copy->pasted the metadata from EN topics

zdrawku:
I have created a separate issue #244 based on your initial comments, I agree with you that metadata translations could be done in separate PR.

jp/components/grid_column_pinning.md
zdrawku: Lines have been added as well as Additional resources section for each new subtopic

ig-ealbert:
Ok, I will work on #244 when this is merged.

PR #288, opened 2018-04-17
fix(build): Fixing failing docfx build

zdrawku:


zdrawku:
@damyanpetev, I agree with you ;)
https://github.com/IgniteUI/igniteui-docfx/issues/289

zdrawku:
@ig-ealbert I have added JP version for this file, should be localized as well.

PR #291, opened 2018-04-17
removed empty space, renamed tooltip sample

pamelabrasil:


gmurray81:
is 550 going to be a bit small for some of the finchart samples in terms of height?

zdrawku:
Adding this tag in order to check it later for localization changes.

PR #292, opened 2018-04-17
adding new sample / doc

pamelabrasil:


zdrawku:
Adding this tag in order to check it later for localization changes.

PR #293, opened 2018-04-18
renamed samples and reordered in TOC

pamelabrasil:


zdrawku:
@ig-ealbert , adding pending-localization in order to be sure that you saw this change

PR #313, opened 2018-04-19
Column resizing demo refactoring

SAndreeva:


ig-ealbert:
I have made the link changes, but the new content in the resizing topic is out of date now.  So I'm leaving the English content in there that was added in a later pull request.

PR #314, opened 2018-04-19
Update docs to match updated samples

simeonoff:


npaunov:
@simeonoff  could you modify in both Radio Button and Switch samples:
-  In API Summary "@Input() labelPosition" there is an additional ` symbol.
-  the formatting/ description of methods (so they are similar to Button Group )

PR #375, opened 2018-04-25
added initial .md files for new 18.2 controls

HUSSAR-mtrela:
added initial .md files for new controls
updated toc files with new topics


ig-ealbert:
@pamelabrasil @HUSSAR-mtrela @yoshimis Please see my review.
In short, adding placeholder files prevents us from pushing new builds to production.  All work should be done in its own branch until it is **completed**.

PR #380, opened 2018-04-26
Update grid topics for date object usage

### LINKED ISSUE #377: Add Note for Grid filtering about string objects when using filtering and sorting
Only date objects should be passed when filter/sort Date type columns. Using of string objects is not allowed.

zdrawku:
Add explanation that additional changes should be made, on application level, for features like sorting and filtering, when it comes to using of `string` values on `Date` column.

Close #377

### DIFF
--- a/en/components/grid_filtering.md
+++ b/en/components/grid_filtering.md
@@ -26,6 +26,9 @@ There's a default filtering strategy provided out of the box, as well as all the
 <igx-column field="Price" filterable="true" dataType="number"></igx-column>
 ```
 
+> [!NOTE]
+> If values of type `string` are used by column of dataType `Date`, the grid won't parse it to `Date` objects and using of filtering conditions won't be possible. If you want to use `string` objects, additional logic should be implemented on application level, in order to parse the values to `Date` object.
+
 You can filter any column or a combination of columns through the grid API. The grid exposes two methods for this task:
 
 *   `filter` - filter a single column or a combination of columns.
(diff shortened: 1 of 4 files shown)


en/components/grid_filtering.md
```diff
@@ -26,6 +26,9 @@ There's a default filtering strategy provided out of the box, as well as all the
 <igx-column field="Price" filterable="true" dataType="number"></igx-column>
 ```
 
+> [!NOTE]
+> If values of type `string` are used by column of dataType `Date`, the grid won't parse it to `Date` objects and using of filtering conditions won't be possible. If you want to use `string` objects, additional logic should be implemented on application level, in order to parse the values to `Date` object.
```
ig-ealbert: ... won't parse it to `Date` objects and using filtering conditions...
or
... won't parse it to `Date` objects and usage of filtering conditions...
I like the first one better, it sounds more natural.

Also, 
... should be implemented on the application level
or
... should be implemented on an application level

en/components/grid_sorting.md
```diff
@@ -68,6 +68,10 @@ public ngOnInit() {
     ];
 }
 ```
+
+> [!NOTE]
+> If values of type `string` are used by column of dataType `Date`, the grid won't parse it to `Date` objects and using of igxGrid `sorting` won't work as expected. If you want to use `string` objects, additional logic should be implemented on application level, in order to parse the values to `Date` object.
```
ig-ealbert: Same comments here.

zdrawku:
@ig-ealbert both en and jp versions are updated for sorting and filtering subtopics

ig-ealbert:
I can add the translations today.

PR #388, opened 2018-04-27
Update grid_selection.md

yoshimis:


jp/components/grid_selection.md
```diff
@@ -7,7 +7,7 @@ _language: ja
 
 ### Data Grid 行選択
 
-Ignite UI for Angular 行選択を使用すると、行ですべての他の列の前に描画されるチェックボックス列があります。
+Ignite UI for Angular 行選択は、行内のすべての列の前に描画されるチェックボックス列にあるチェックボックスで行を選択/選択解除でき、データを含む複数の行を選択/選択解除できます。
```
ig-ealbert: This was also content I was going to add from the .sdlxliff file, but I'll assume this is more up to date and won't overwrite it.

PR #406, opened 2018-05-14
chore(igxGrid): add desc about newly exposed dblclick evt

Aleksandyr:
Related to [1201](https://github.com/IgniteUI/igniteui-angular/issues/1201) issue.


zdrawku:
Thank you @ig-ealbert, @yoshimis once you are ready with the translation, could you remove the label with status:pending-localization and add label with status:localized

PR #415, opened 2018-05-29
docs(grid): Add "searchable" property to grid-search doc

tachojelev:
docs(grid): Add "searchable" property to grid-search doc

IgniteUI/igniteui-angular#1400


ig-ealbert:
@yoshimis This should be ready if you can update the branch.

PR #449, opened 2018-06-18
chore(igxDatePicker): document new inputs

Aleksandyr:


kdinev:
@ig-ealbert It's misspelled in the document. I will fix it right away.

PR #467, opened 2018-06-25
docs(grid): Updated docs for filtering with latest changes #1355

sboykova:
Related to IgniteUI/igniteui-angular#1355

### DIFF
--- a/en/components/grid_filtering.md
+++ b/en/components/grid_filtering.md
@@ -99,7 +99,7 @@ this.grid.clearFilter();
 
 #### Initial filtered state
 
-It is possible to set the initial filtering state of the grid by an array of expressions to the `filteringExpressionsTree` property of the grid.
+To set the initial filtering state of the grid, set the IgxGridComponent `filteringExpressionsTree` property to an array of IFilteringExpressionsTree for each column to be filtered.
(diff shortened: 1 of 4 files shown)


en/components/grid_filtering.md
```diff
@@ -73,14 +99,22 @@ this.grid.clearFilter();
 
 #### Initial filtered state
 
-It is possible to set the initial filtering state of the grid by an array of expressions to the `filteringExpressions` property of the grid.
+It is possible to set the initial filtering state of the grid by an array of expressions to the `filteringExpressionsTree` property of the grid.
```
bkulov: You could remove the `initial`. The user could use this property to set grid's filtering state at any time.

bkulov: Also this sentence doesn't correspond to the sample bellow. The sentence says `filteringExpressionsTree` should contain an array of expressions which is not true. It should contain an array of `IFilteringExpressionsTree` for each column the user wants to filter.

yoshimis:
Please update codes in grid_filtering for JP.

kdinev:
@sboykova Please apply the code updates to the JP version.

PR #491, opened 2018-06-27
docs(groupby): Fix GroupBy sample dimensions.

mpavlinov:


ig-ealbert:
@yoshimis @mpavlinov This should be fixed in Japanese too.

ig-ealbert:
I see this is being done in #494.

mpavlinov:
@ig-ealbert We have to revert #494, because the dropdowns are not visible. Will make another PR.

PR #506, opened 2018-06-28
docs(grid): Fix in column moving demo - added comments for pinning

sboykova:


ig-ealbert:
Looks like this should be tagged as pending-localization?

PR #519, opened 2018-06-29
Ys 0629

yoshimis:


jp/components/drop_down.md
```diff
@@ -286,9 +286,9 @@ export class AppModule {}
 
 | 名前 | 型 | 説明 |
 | :--- | :--- | :--- |
-| `isHeader` | ブール値| 項目がグループ ヘッダーかどうかを定義します。 |
+| `isHeader` | boolean| 項目がグループ ヘッダーかどうかを定義します。 |
 | `disabled` | boolean| 項目が無効かどうかを取得します。|
```
kdinev: Shouldn't this be translated as well?

yoshimis: It was my mistake to translate type. it is good as boolean.

PR #520, opened 2018-06-29
Update iframe id

zdrawku:


kdinev:
@zdrawku Does this need a JP update as well?

zdrawku:
Yes @kdinev, thank you!

PR #522, opened 2018-06-29
Add all operand to boolean operands

sstoyanovIG:
IgniteUI/igniteui-angular#1773


ig-ealbert:
I added the translation.  In general, if we're going to ship in < 1 day, please do not add any English to the Japanese topics.  We'll get them in the next update.

PR #536, opened 2018-07-04
docs(themes): add more topics related to theming

simeonoff:


ig-ealbert:
When the Japanese translations are added, the docfx.json file must also be updated.

yoshimis:
Asked Simeon to replace EN files with JP ones.

ig-ealbert:
I will add the last two files in Japanese as well as make fixes.  I will update the docfx.json file as I had requested be done.

PR #562, opened 2018-07-11
Updated gauges topics

ig-ealbert:


ig-ealbert:
@pamelabrasil Could you please merge these changes if you approve?  Then we can start testing.

PR #569, opened 2018-07-13
Matched changes from PR 565

ig-ealbert:
#565 
Note that some content was removed in English.  I'm matching those changes, assuming it was intentional.


HUSSAR-mtrela:
yes, that content was removed intentionally. 
FYI, We will be adding more samples to these 3 gauge topics because Jason requested it.

ig-ealbert:
Could someone please merge this?

PR #592, opened 2018-07-23
Updated for #586

yoshimis:


zdrawku:
@HristoP96, could you verify that the project is loading okay

PR #602, opened 2018-07-25
Refactoring the "Export to Excel" topic adding paste from Excel section

mpavlinov:
The "Export to Excel" topic is now renamed to "Integration with Excel" and includes the "Paste from Excel" section.


mpavlinov:
@MayaKirova Please add more description to the `paste-handler` directive section if think it's needed.

PR #628, opened 2018-08-06
Content Updates to TOC and Grid Overview Topic

jasonzajac:
Updated trial link to point to "https://www.infragistics.com/products/ignite-ui-angular/download" within en/toc.yml and en/components/grids_and_lists.tmpl. Also made some other minor content changes to grids and lists overview topic.

**Please note**: for localization only, do not change the free trial link for JP until @DaizenIG confirms that this will be used for that instance.


yoshimis:
@jasonzajac please add 'pending-localization' label if there is any further changes and ready to localize.

jasonzajac:
@mpavlinov and @yoshimis apologies for the back and forth with this one. I had to confirm a few things with @DaizenIG .

JP is not ready to change the Download link at the moment, but they can keep the changes made to the Grids and Lists topic. Specifically, I reverted some changes Yoshimi made related to the link, but kept the Column hiding update.

Once you both reviewed, let me know if it will be ok to proceed the changes in the current state. I will create a new PR when JP is ready to go live with their new download link.

PR #667, opened 2018-08-27
docs(tabs): Fixed stackbliz links #1957

sboykova:
Related to IgniteUI/igniteui-angular#1957


zdrawku:
We've decided to remove the embed samples from this topic, because of routing issues with the embed functionallity. Stackblitz samples are working just fine though, which is enough.

PR #679, opened 2018-09-06
fix category chart Import

igDivya:
Bug Fix: TFS257057


ig-ealbert:
@pamelabrasil I have added the pending-localization label.  Please add this label when the your team's changes are approved.  Yoshimi filters based on this tag to find what needs to be done.  Alternatively, if it's a change that doesn't require any translation, you can make it in both languages.

PR #683, opened 2018-09-12
Update allowItemsFocus usage and description

Lipata:


en/components/drop_down.md
```diff
@@ -224,6 +224,38 @@ If the sample is configured properly, a list  of countries should be displayed a
     </button>
 </div>
 
+### igxDropDownItemNavigation directive
+
+To enable keyboard navigation for the igxDropDown component, the igxDropDownItemNavigation directive should be applied. It's recommended that the directive is applied to the element that is triggering the opening and closing the of the drop down. That element will also have the focus, and when using the directive it will handle all triggered events. By default the focus will always stay on the element that igxDropDownItemNavigation is applied on, and this happens because the igxDropDown doesn't focus its items by default.
```
wnvko: It's recommended that the directive is applied to the element that is triggering the opening and closing **the of the** drop down - this should be corrected.

en/components/drop_down.md
```diff
@@ -224,6 +224,38 @@ If the sample is configured properly, a list  of countries should be displayed a
...
+    <igx-suffix><igx-icon [name]="arrowProvince"></igx-icon></igx-suffix>
+</igx-input-group>
+<igx-drop-down #dropdownProvince>
+    <igx-drop-down-item *ngFor="let p of provinceData">
+        {{ p }}
+    </igx-drop-down-item>
+</igx-drop-down>
+```
+
+If you want the drop down items to receive focus then you need to explicitly set `allowItemsFocus` property of the igxDropDown to `true`. In that case you can apply the igxDropDownItemNavigation on the drop down itself, however this can also happen on the input again. Applying the directive on the drop down means that you can omit the value, because it is the drop down itself.
```
wnvko: Applying the directive on the drop down means that you can omit the value, because it is the drop down itself
This is not clear enough. Should we change value with reference?

jp/components/drop_down.md
```diff
@@ -221,6 +221,38 @@ export class AppModule {}
...
+    <igx-suffix><igx-icon [name]="arrowProvince"></igx-icon></igx-suffix>
+</igx-input-group>
+<igx-drop-down #dropdownProvince>
+    <igx-drop-down-item *ngFor="let p of provinceData">
+        {{ p }}
+    </igx-drop-down-item>
+</igx-drop-down>
+```
+
+If you want the drop down items to receive focus then you need to explicitly set `allowItemsFocus` property of the igxDropDown to `true`. In that case you can apply the igxDropDownItemNavigation on the drop down itself, however this can also happen on the input again. Applying the directive on the drop down means that you can omit the value, because it is the drop down itself.
```
wnvko: Applying the directive on the drop down means that you can omit the value, because it is the drop down itself.
Same as above

en/components/drop_down.md
damyanpetev: "To enable keyboard navigation for the igxDropDown component, the igxDropDownItemNavigation directive ~should~**can** be applied. ~It's recommended that t~**T**he directive ~is~**should be** applied to the **active (focused)** element **or a parent container** ~is triggering the opening and closing the of the drop down~. ~That element will also have the focus, and when using the directive it will~**This will allow the directive to** handle all triggered events. By default ~the focus will always stay on the element that igxDropDownItemNavigation is applied on, and this happens because~ the igxDropDown doesn't **take** focus ~its items by default.~ **so for example the directive can be placed on a button or input that controls the drop down. When the drop down's `allowItemsFocus` is enabled the directive can be placed either on the component itself or on each item.**"

damyanpetev: TBH, this bit should also list the actual keyboard navigation the directive provides or link to one of the specs that describe it.

en/components/drop_down.md
```diff
@@ -224,6 +224,38 @@ If the sample is configured properly, a list  of countries should be displayed a
     </button>
 </div>
 
+### igxDropDownItemNavigation directive
+
+To enable keyboard navigation for the igxDropDown component, the igxDropDownItemNavigation directive should be applied. It's recommended that the directive is applied to the element that is triggering the opening and closing the of the drop down. That element will also have the focus, and when using the directive it will handle all triggered events. By default the focus will always stay on the element that igxDropDownItemNavigation is applied on, and this happens because the igxDropDown doesn't focus its items by default.
+The following sample demonstrates an input that on click opens and closes igxDropDown instance. Applying the igxDropDownItemNavigation on the input itself, will enable keyboard navigation, when using arrow up and arrow down, and in the same the input will not loose the focus during that process.
```
damyanpetev: This example should be simplified - just an input + directive that can still open the drop down on arrow down and continue navigating afterwards.

damyanpetev: \+ "This relies on the default drop down behavior with `allowItemsFocus` disabled to allow the input to maintain focus"

en/components/drop_down.md
damyanpetev: target?

damyanpetev: ~If you want the drop down items to receive focus then you need to explicitly set~**When the** `allowItemsFocus` property of the igxDropDown **is set** to `true`~. In that case~ you can apply the igxDropDownItemNavigation on the drop down itself **or the items**~, however this can also happen on the input again~. Applying the directive on the drop down means that you can omit the value, because it is the drop down itself.

en/components/drop_down.md
```diff
@@ -226,30 +226,52 @@ If the sample is configured properly, a list  of countries should be displayed a
...
-The following sample demonstrates an input that on click opens and closes igxDropDown instance. Applying the igxDropDownItemNavigation on the input itself, will enable keyboard navigation, when using arrow up and arrow down, and in the same the input will not loose the focus during that process.
+To enable keyboard navigation for the `igxDropDown` component, the `igxDropDownItemNavigation` directive can be applied. The directive should be applied to the active(focused) or a parent container. This will allow the directive to handle all triggered events. By default the the igxDropDown or its items don't take focus, so for example the directive can be placed on a button or input that controls the drop down.
```
damyanpetev: The directive should be applied to the active(focused) **element**

en/components/drop_down.md
```diff
@@ -226,30 +226,52 @@ If the sample is configured properly, a list  of countries should be displayed a
...
+The `igxDropDownItemNavigation` directive value should be target component that is or extends `IgxDropDownBase` class. This ensures that the target component implements the following methods:
+
+| Name | Description |
```
damyanpetev: Not sure this is needed, you can probably just link to the API page?

yoshimis:
en/components/drop_down.md contains same description for `Enter` and `Space`. Is that correct?

Lipata:
Yes, that's right. Thanks!

jp/components/drop_down.md
```diff
@@ -221,6 +221,56 @@ export class AppModule {}
     </button>
 </div>
 
+### igxDropDownItemNavigation ディレクティブ
+
+igxDropDownItemNavigation ディレクティブ
```
ig-ealbert: @yoshimis I don't think this needs to be here?  Looks like a copy of the header without the `###`?

jp/components/drop_down.md
```diff
@@ -221,6 +221,56 @@ export class AppModule {}
...
+`igxDropDown` コンポーネントのキーボード ナビゲーションを有効にするには、`igxDropDownItemNavigation` ディレクティブを適用します。ディレクティブをアクティブ (フォーカスした) 要素または親コンテナに適用することにより、トリガーしたすべてのイベントを処理できるようになります。デフォルトでは、igxDropDown またはその項目にフォーカスが当たらないため、たとえばディレクティブをドロップダウンを制御するボタンや入力に配置できます。
+`igxDropDownItemNavigation` ディレクティブ値は、ターゲット コンポーネントまたは拡張 `IgxDropDownBase` クラスにする必要があります。
+
+
+以下のサンプルは、 on click オープンおよびクローズ の igxDropDown インスタンスの例です。入力に `igxDropDownItemNavigation` を適用し、上矢印と下矢印の使用時にキーボード ナビゲーションを有効にします。`allowItemsFocus` が無効のデフォルトのドロップダウン動作に依存し、ボタン入力にフォーカスを維持することが可能です。
```
ig-ealbert: "on click" here means "when clicked", so we should probably translate that.

jp/components/drop_down.md
```diff
@@ -221,6 +221,56 @@ export class AppModule {}
...
+    <input igxInput type="text">
+</igx-input-group>
+<igx-drop-down #dropdownProvince [allowItemsFocus]="true">
+    <igx-drop-down-item *ngFor="let p of provinceData" [igxDropDownItemNavigation]="dropdownProvince">
+        {{ p }}
+    </igx-drop-down-item>
+</igx-drop-down>
+```
+
+<div class="divider--half"></div>
```
ig-ealbert: This line isn't necessary either.

PR #692, opened 2018-09-20
Add limitations table to the docs

### LINKED ISSUE #691: Add limitations table to the Grid Search dos
The grid search docs should include the known search limitations

DDavidkov:
Closes #691


en/components/grid_search.md
```diff
@@ -287,6 +287,15 @@ On the right in our input group, let's create three separate containers with the
 </igx-suffix>
 ```
 
+## Known Limitations
+
+|Limitation|Description|
+|--- |--- |
+|Searching in cells with a template|The search functionality will replace any cell template with a div element and spans which highlight the searched text in the cell value. This means that it can lead to unexpected behavior when you have more complicated templates, containing several elements or events hooked into the template. If you have a column with such template, you should either use alternative approaches, such as a column formatter, or to set the `searchable` property on the column to false.|
```
gedinakova: "or **to** set the `searchable` property" - no "to".

PR #711, opened 2018-09-27
Update overlay topic API links

PlamenaMiteva:


PlamenaMiteva:
I have changed the environment link as @zdrawku has requested

en/components/overlay_main.md
```diff
@@ -249,92 +249,17 @@ export class ExampleComponent {
 
 ## API
```
zdrawku: Here we should list only `components` API, for example you can check the Grid editing topic:

https://github.com/IgniteUI/igniteui-docfx/blob/master/en/components/grid_editing.md#api

The topic is referencing GridCellComponent and GridRowComponent mainly, this is why the API section is listing only them.

All other services, properties and interfaces that are used, are not listed.

PlamenaMiteva: I have left the link to the overlay service itself and removed all the other links.

Issue #702, opened 2018-09-27
Typedoc references in topics


zdrawku:
For all topics related to the features/components below we should add a Typedoc references.

When some method or property in mentioned in the topic, the corresponding link to the Typedoc should be provided.

At the end of each topic add `API Reference` section and there list all the APIs that are mentioned in the topic, example:

```
## API Reference
[IgxExpansionPanel API]({environment:infragisticsBaseUrl}/products/ignite-ui-angular/docs/typescript/classes/igxexpansionpanelcomponent.html)
```

- [x] components/grid.html
- [x] components/grid_editing.html
- [x] components/grid_paging.html
- [x] components/grid_sorting.html
- [x] components/grid_summaries.html
- [x] components/grid_multi_column_headers.html
- [x] components/grid_displaydensity.html
- [x] components/linear_progress.html
- [x] components/circular_progress.html
- [x] components/dialog.html
- [x] components/slider.html
- [x] components/toggle.html
- [x] components/layout.html
- [x] components/carousel.html
- [x] components/calendar.html
- [x] components/date_picker.html

We are going to update the issue with listing of all topics based on the ownership.

PR #716, opened 2018-09-27
Main grid changes

zdrawku:
Related to #702


en/components/grid.md
```diff
@@ -22,7 +22,7 @@ _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI
 
 ### Dependencies
 
-The grid is exported as an `NgModule`, thus all you need to do in your application is to import the _IgxGridModule_ inside your `AppModule`:
+The grid is exported as an `NgModule`, thus all you need to do in your application is to import the [_IgxGridModule_]({environment:angularApiUrl}/classes/igxgridmodule.html) inside your `AppModule`:
```
hanastasov: why dont you wrap the name in `` ?

en/components/grid.md
```diff
@@ -177,7 +177,7 @@ const column = this.grid.getColumnByName("User");
 column.bodyTemplate = this.smallView;
 ```
 
-Column properties can also be set in code in the **initColumns** event which is emitted when the columns are initialized in the grid.
+Column properties can also be set in code in the [`initColumns`]({environment:angularApiUrl}/classes/igxgridcomponent.html#oncolumninit) event which is emitted when the columns are initialized in the grid.
```
hanastasov: event is called onColumnsInit actually.

hanastasov: resolved in another PR

PR #728, opened 2018-10-02
Added topic for drag and drop.

MartinKamenov:


en/components/drag_drop.md
```diff
@@ -0,0 +1,95 @@
...
+<div igxDrag>Drag me</div>
+```
+
+#### DOM Behavior
+
+By default a drag operation starts when the end user swipes at least 5 px in any direction. Otherwise the interaction is considered as a click and the [`dragClicked`]({environment:angularApiUrl}/classes/igxdragdirective.html#dragclicked) event is emitted.
+
+When dragging occurs a drag ghost element is spawned and moves along with the mouse cursor or touch interaction. The original element is still present, but it can be hidden automatically when dragging starts with the [`hideBaseOnDrag`]({environment:angularApiUrl}/classes/igxdragdirective.html#hidebaseondrag) input.
+
+The dragging can be canceled by setting the `cancel` property of the [`dragStart`]({environment:angularApiUrl}/classes/igxdragdirective.html#dragstart) event to `true`. This will cancel the default dragging logic.
```
HristoP96: Provide reference to the `cancel`  property of the IDragStartEventArgs interface:
[`cancel`] ({environment:angularApiUrl}/interfaces/idragstarteventargs.html#cancel)

en/components/drag_drop.md
```diff
@@ -0,0 +1,96 @@
...
+<div igxDrag [hideBaseOnDrag]="true" [animateOnRelease]="true" *ngFor="let elem of draggableElems" >
+    <span [style.margin]="'auto'">{{elem.label}}</span>
+</div>
+```
+
+### Drop Directive
+
+When an element that is being dragged using the [`igxDrag`]({environment:angularApiUrl}/classes/igxdragdirective.html) directive needs to be placed in an area, the [`igxDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html) can be used to achieve this behavior. It provides events that the user can use to determine if element is entering the drop area and if it is being released inside it.
+
+#### Basic configuration
```
skrustev: Could you add more information regarding the default behavior of the `igxDrop` and how the user can cancel and override that default behavior.
You can see the readme for more information regarding that.

en/components/drag_drop.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Basic Configuration
+The [`igxDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html) directive can be applied to any DOM element just like the [`igxDrag`]({environment:angularApiUrl}/classes/igxdragdirective.html) directive. 
+
+````html
+<div igxDrop>Drop here</div>
+````
+
+By default the [`igxDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html) directive comes with logic that appends the dropped [`igxDrag`]({environment:angularApiUrl}/classes/igxdragdirective.html) element as a child of the element that has instanced the [`igxDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html). It can be overridden by canceling the [`onDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html#ondrop) event of the [`igxDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html) directive. This can be done by setting the `cancel` argument that the [`onDrop`]({environment:angularApiUrl}/classes/igxdropdirective.html#ondrop) event provides.
```
skrustev: I noticed that the `cancel` argument above for the `dragStart` has been linked also if we want to keep it consistent.

PR #737, opened 2018-10-10
Add row editing and transaction topics

### LINKED ISSUE #736: Create topics for row editing and transactions

Lipata:
closes #736


PlamenaMiteva:
Some minor spelling typos:
1. On the grid-editing page just below the note the 'key' word is missing in the sentence:

> You can also modify the cell value through the IgxGrid API but only if primary **key** is defined:

2. A spelling typo on the grid-row-editing page in the Navigation section:

> If the new cell **thay** gets focused is editable, then the new row also enters edit mode, while if the cell is not editable, then only the previous row exits edit mode.

3. On the grid-transactions page the last section should be 'Additional Resources' instead of 'Methods' and the links are not correct

PR #744, opened 2018-10-12
updating for domain chart API changes

mharrington622:
I tried to update the JP pages too, but couldn't get the encoding right.


mharrington622:
Actually, just going to delete this and redo it.  I left some orphaned commits lying around.

mharrington622:
New PR: #745

Issue #749, opened 2018-10-17
Korean language support


tiliev:
Please implement Korean language support. Tasks:
- [x] 1. Introduce new subdomain for the [igniteui-angular-samples](https://github.com/IgniteUI/igniteui-angular-samples). Staging environment - https://kr.staging.infragistics.local/angular-demos/ and production environment - https://kr.infragistics.com/angular-demos/.
- [x] 2. Update [igniteui-docfx-template](https://github.com/IgniteUI/igniteui-docfx-template)'s environment variables [file](https://github.com/IgniteUI/igniteui-docfx-template/blob/master/post-processors/PostProcessors/EnvironmentVariables/preconfig.json). (https://github.com/IgniteUI/igniteui-docfx-template/issues/45)
- [x] 3. Introduce [igniteui-docfx-template](https://github.com/IgniteUI/igniteui-docfx-template)'s [resources](https://github.com/IgniteUI/igniteui-docfx-template/blob/master/resources/resources-en.json) in Korean. (https://github.com/IgniteUI/igniteui-docfx-template/issues/45)
- [x] 4. Localize [igniteui-docfx-template](https://github.com/IgniteUI/igniteui-docfx-template)'s [resources](https://github.com/IgniteUI/igniteui-docfx-template/blob/master/resources/resources-en.json) in Korean. (https://github.com/IgniteUI/igniteui-docfx-template/issues/45)
- [x] 5. Introduce new Google Analytics Containers for the Korean version of the documentation for development, staging and production environments. (https://github.com/IgniteUI/igniteui-docfx-template/issues/45)
- [x] 6. Localize Ignite UI DocFX [resources](https://github.com/IgniteUI/igniteui-docfx/blob/master/en/global.json).
- [x] 7. Introduce a new folder named `kr` in the current repository. It should contain the `en` folder contents. Update DocFX build logic. Update repository read me.
- [x] 8. Integrate the localized Korean header & footer from infragistics.com. Expected staging URL - https://kr.staging.infragistics.local/navigation, production - https://kr.infragistics.com/navigation. (https://github.com/IgniteUI/igniteui-docfx-template/issues/45)
- [x] 9. Introduce a new subdomain for the current repository supporting Korean. Staging environment - https://kr.staging.infragistics.local/products/ignite-ui-angular/angular/. Production environment - https://kr.infragistics.com/products/ignite-ui-angular/angular/.
- [x] 10. Translate grid topics.

tiliev:
[...]

PR #759, opened 2018-10-23
Korean language support

yoshimis:
Related to https://github.com/IgniteUI/igniteui-docfx/issues/749.


zdrawku:
Also new docfx-template version should be tagged, and all build warning should be fixed.

PR #763, opened 2018-10-29
topics for excel, pie and doughnut charts

HUSSAR-mtrela:
image

image


mddifilippo89:
Can we change these headers to match WPF: https://staging.infragistics.local/help/wpf/excelengine-using-the-infragistics-excel-engine

PR #766, opened 2018-10-30
updating pie and doughnut

pamelabrasil:


yoshimis:
@pamelabrasil could you please update code section in JP and KR next time?

PR #838, opened 2018-11-13
removed staging links

pamelabrasil:


zdrawku:
@pamelabrasil we are going to fix all environment references, thank you for noticing them.

Issue #841, opened 2018-11-19
Drop down samples truncated.  


IvayloG:
- Countries drop down sample is truncated on bottom.
- Left box-shadow is not visible for all samples.
- Items lists and buttons with different widths.

Related to https://github.com/IgniteUI/igniteui-angular-samples/issues/759

PR #842, opened 2018-11-19
Fix truncated drop down samples.

IvayloG:
#841


damyanpetev:
@IvayloG On second thought, can you also apply the same change to jp/kr files as well :)

PR #844, opened 2018-11-19
Check-in for Excel doc bugs.

agoldenbaum:
For TFS items:
259405,
259408,
259457

- Regarding the cell operations topic - there is a lot changed or removed here. Many of the things removed are either not implemented in the Angular tool set or they are not really "cell operations." For example, adding/removing columns and rows is part of the Worksheet Operations topic.


yoshimis:
Updated as #840.

PR #845, opened 2018-11-20
Kr 1120

yoshimis:


yoshimis:
@zdrawku Yes, that is correct. Till next KR update scheduled for next year, any new pages should be added as-is (In English).

PR #870, opened 2018-12-04
Setting bigger height to calendar samples iframe for prevention from cutting.

HristoP96:
The samples "calendar-sample-1" and "calendar-sample-5" need higher iframe because of their new positioning from this [PR](https://github.com/IgniteUI/igniteui-angular-samples/pull/783).
Otherwise the calendars will be cut.


HristoP96:
@hanastasov If all of the calenders are expanded, due to month selection, the blank space will be filled.

hanastasov:
> @hanastasov If all of the calenders are expanded, due to month selection, the blank space will be filled.

Then we better keep the calendar in one line, so their expanding does not adds up the requireed space below.

PR #879, opened 2018-12-07
Filtering operands update

kdinev:


yoshimis:
KR should be localized at next outsource.

PR #900, opened 2018-12-12
Add timepicker dropdown sample.

npaunov:
Related IgniteUI/igniteui-angular#3394


en/components/time_picker.md
```diff
@@ -151,9 +151,47 @@ And there we have it:
...
+```html
+<!--timePickerDropdown.component.html-->
+
+<igx-time-picker [mode]="mode"></igx-time-picker>
+```
+The user now will be able to type, edit or delete the time value in the input in both 12- and 24-hour formats. 
+
+**Dropdown Mode with Keyboard Navigation**
+
+The mouse caret can be positioned at the hours, minutes or AM/PM placeholders and after pressing <kbd>Up</kbd> or using `Mouse Wheel Up`, the hours or the minutes will be increased. Pressing <kbd>Down</kbd> or `Mouse Wheel Down` can be used for the reversed operation. *Note* that if the time picker's `minValue` or `maxValue` are set and `isspinloop` is disabled, the time scrolling will break at the specified hour or minute value. In addition, the user can also perform the following operations:
```
SAndreeva: minValue and maxValue should be provided as links. 
Can you correct isspinloop to `isSpinLoop` and provide a linik for it as well.

Issue #921, opened 2019-01-07
Update new sample URLs


tiliev:
Since we are making lazy loading, all of the sample URLs will be changed. The bundles are as follows:
## Bundles
- [x] **Grid CRM** bundle
- [x] **Charts** bundle
- [x] **Gauges** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/928)
- [x] **Excel library** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/937)
- [x] **Grids** bundle - Data Grid + TreeGrid
- [x] **Lists** bundle - List View + Combo
- [x] **Styling and themes** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/934)
- [x] **Data Entry** bundle
- [x] **Data Display** bundle
- [x] **Interactions** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/927)
- [x] **Menus** bundle
- [x] **Layouts** bundle
- [x] **Notifications** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/922)
- [x] **Scheduling** bundle (https://github.com/IgniteUI/igniteui-docfx/pull/931)
- [x] **Services** bundle

PR #922, opened 2019-01-07
Updated notifications samples with the new URLs

tiliev:
Related to #921.

For the following components:
 - Banner
- Snackbar
- Toast

please test whether the samples in the corresponding topics open and whether the same samples work in StackBlitz.


hanastasov:
npm install resolves the issue.

PR #955, opened 2019-01-18
localization for JP

yoshimis:


yoshimis:
@zdrawku It is there.
image

PR #976, opened 2019-01-25
Fix issue 975

### LINKED ISSUE #975: Grid's Boston marathon sample is not loaded correctly
Open grid topic of KR staging

Error is thrown:
image

The routh should point to grid/grid

hanastasov:
Closes #975


yoshimis:
KR will be outsourced and translated at 19.1 release timing (probably on April).

PR #1003, opened 2019-02-06
Changes from AngularDocFX_EN

ESShared:


tiliev:
@HUSSAR-mtrela, I noticed the Excel library samples have their URLs updated. The prefix `excel-library` is removed. Is this intentional?
Also we have a lot of TODOs, but I guess these are going to be updated before the PR gets merged.
Thanks.

HUSSAR-mtrela:
@tiliev we added missing "excel-library" prefix in excel topic

PR #1013, opened 2019-02-12
refactor(button): add documentation for the outlined button style

simeonoff:


StefanIvanov:
@yoshimis could you please update us when do you plan to have the JP for us

PR #1018, opened 2019-02-13
Update redirecting URLs to true source for SEO boost

jasonzajac:
Updated page links that were redirecting to use the true source of the resource. This is for our overall SEO effort.


tiliev:
@jasonzajac, please resolve the conflicts and I'll take a look. Thank you.

PR #1028, opened 2019-02-18
Add IgxSelect topic

jackofdiamond5:
New samples:
- Basic select sample.
- Select with groups sample.
- Select in a form sample.
- Select with custom overlay settings sample.

### DIFF
--- a/en/components/select.md
+++ b/en/components/select.md
@@ -5,7 +5,7 @@ _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI 
 ---
 
 ## Select
-The `IgxSelectComponent` allows you to select a single item from a drop-down list, by using the mouse or the keyboard to quickly navigate through them. Using the `igxSelect` you can also iterate selection through all items based on the input of a specific character or multiple characters.
+The `IgxSelectComponent` allows you to select a single item from a drop-down list by using the mouse or the keyboard to quickly navigate through them. Using `igx-select` you can also iterate selection through all items based on the input of a specific character or multiple characters.
 
 ## Simple Select Demo
 <div class="sample-container loading" style="height: 330px;">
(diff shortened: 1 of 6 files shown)


en/components/select.md
```diff
@@ -0,0 +1,370 @@
...
+
+## Simple Select Demo
+<div class="sample-container loading" style="height: 330px;">
+    <iframe id="select-sample-1-iframe" frameborder="0" seamless width="100%" height="100%" src="{environment:demosBaseUrl}/data-entries/select-sample-1" onload="onSampleIframeContentLoaded(this);"></iframe>
+</div>
+
+<button data-localize="stackblitz" disabled class="stackblitz-btn" data-iframe-id="dropdown-sample-1-iframe" data-demos-base-url="{environment:demosBaseUrl}">view on stackblitz</button>
+
+>[!WARNING]
+>To start using Ignite UI for Angular components, in your own projects, make sure you have configured all necessary dependencies and have performed the proper setup of your project. You can learn how to do this in the [*installation*](https://www.infragistics.com/products/ignite-ui-angular/getting-started#installation) topic.
```
bazal4o: Let's edit the link title to Getting started topic - instead of installation

en/components/select.md
```diff
@@ -0,0 +1,370 @@
...
+
+You can make use of the `onClosed` event like so:
+```html
+<igx-select (onClosed)="handleClosed($event)">
+    <igx-select-item [value]="Apple">Apple</igx-select-item>
+</igx-select>
+```
+
+You put all your *handler* functions inside of your *class*:
+```ts
```
bazal4o: Enhance the example with at least one event canceled

en/components/select.md
```diff
@@ -0,0 +1,370 @@
...
+```
+
+#### Opening/Closing events
+- Emitted on:
+    - input click
+    - select expand/collapse button click (app scenario)
+- Triggered on:
+    - key interaction
+
+The thing about the `onOpening` and `onClosing` events is that they are fired *before* the animation finishes playing, i.e. before the drop-down is fully **opened** or **closed**.
```
bazal4o: The thing is not really correct wording. Please reword. One other difference should be, those events are cancel-able I think - if so please mention it

en/components/select.md
```diff
@@ -0,0 +1,370 @@
...
+        // return something
+    }
+    /* --- */
+}
+```
+- Please note that the above examples are for demonstration purposes only and are not meant to abide by any code standards.
+
+### Positioning Strategy
+`igx-select` has its own positioning strategy called the `SelectPositioningStrategy`.
+It extends the [*ConnectedPositioningStrategy*](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/classes/connectedpositioningstrategy.html) and allows `igx-select` to position its drop-down list in different ways, relative to the input field.
```
bazal4o: It will be good to mention what the real positioning of the list is (the selected item matches the text into the input)

en/components/select.md
```diff
@@ -0,0 +1,370 @@
...
+
+In order to use the `SelectPositioningStrategy`, you first have to import it alongside the [*OverlaySettings*](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/interfaces/overlaysettings.html) which are going to use it:
```
bazal4o: It's not correct to say in order to use, because the strategy is used by default from the select. Here we configure custom settings which extent the SelectPositionStrategy

PR #1034, opened 2019-02-20
fix(docs): Removing virtualization known issues #1032

### LINKED ISSUE #1032: There are some virtualization limitations that are no longer valid
Those two are fixed already:
*   Variable row heights are not supported. All rows must have the same height.
*   Browsers currently have height limitation for DOM elements. Because of that the total height of the rows shouldn't exceed the height limitation of the browsers. Otherwise there may be unexpected changes in behavior of [`IgxGrid`]({environment:angularApiUrl}/classes/igxgridcomponent.html). For example Internet Explorer 11 has height limitation of 1 533 916 pixels, which for rows with height 50px means no more than 30 678 rows.

Expected result:
Remove those limitations from the docs.

mpavlinov:
Closes #1032


yoshimis:
It appears to be the same changes with #1033. No changes needed for JP and KR.

PR #1035, opened 2019-02-21
Add Hierarchical Grid Load on Demand topic.

skrustev:


en/components/toc.yml
```diff
@@ -198,6 +198,9 @@
     - name: Conditional Cell Styling
       href: hierarchicalGrid/conditional_cell_styling.md
       new: true
+    - name: Load on Demand
+      href: hierarchicalGrid/hierarchical_grid_load_on_demand.md
```
tachojelev: Currently all feature topics for the respective grids contain only the name of the feature (the respective grid is implied by the folder name), so I believe we can rename the md file
from **hierarchical_grid_load_on_demand.md**
to **load_on_demand.md**.

en/components/hierarchicalGrid/hierarchical_grid_load_on_demand.md
```diff
@@ -0,0 +1,305 @@
+---
+title: Hierarchical Grid load on demand - Native Angular | Ignite UI for Angular
+_description: The Ignite UI for Angular Hierarchical Grid provides the necessary tools to load data on demand for each child grid that is expanded. That way the volume of data would be greatly reduced and can be retrieved only when the user needs it.
+_keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI widgets, Angular, Native Angular Components Suite, Native Angular Controls, Native Angular Components Library, Angular Tree Grid component, Angular Tree Grid control, Angular Tree Grid component, Angular High Performance Tree Grid, Hierarchical Grid
```
tachojelev: Keywords should contain **Hierarchical Grid** instead of **Tree Grid**.

en/components/hierarchicalGrid/hierarchical_grid_load_on_demand.md
```diff
@@ -0,0 +1,305 @@
+---
+title: Hierarchical Grid load on demand - Native Angular | Ignite UI for Angular
+_description: The Ignite UI for Angular Hierarchical Grid provides the necessary tools to load data on demand for each child grid that is expanded. That way the volume of data would be greatly reduced and can be retrieved only when the user needs it.
+_keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI widgets, Angular, Native Angular Components Suite, Native Angular Controls, Native Angular Components Library, Angular Tree Grid component, Angular Tree Grid control, Angular Tree Grid component, Angular High Performance Tree Grid, Hierarchical Grid 
+---
+
+## Hierarchical Grid Load On Demand
```
tachojelev: All grid/treegrid/hierarchicalgrid feature topics begin with header 3 (**###**) instead of header 2 (**##**), so for consistency it would be nice to have it that way here as well.

skrustev:
> @skrustev Can you add one paragraph/sentence explaining that expand/collapse buttons will be always visible no matter whether there is data or not on the server. If not then a message will appear after the user expands the grid.

@mpavlinov I think this is a more general information and should be put in the Overview because it is not specific to the Load on Demand but in general when there is row island defined. 
It is mentioned though that the data from the service can be empty when setting the loading indicator.

PR #1093, opened 2019-03-07
Vkacheshmarova/fixing issue 1092

### LINKED ISSUE #1092: There are links to the igxHierarachicalGrid Search topic, which is not available at this point, in some of the igxHGrid topics
In the following topics a link to the  "Hierarchical Grid Searching" can be found in the API References section:
Hierarchical Grid Editing
Hierarchical Grid Column Moving
Hierarchical Grid Column Hiding
Hierarchical Grid Display Density
At this point the search functionality is not yet implemented for igxHGrid, these links should be removed. I will only comment them since the feature will soon be ready and we will bring them back.

kacheshmarova:
Closes #1092 
I am adding 'if' clause for Search topic in igxHGrid topics. When the feature is ready and respectively its topic this 'if' clause will be removed.


en/components/grids_templates/column_hiding.md
```diff
@@ -492,7 +492,9 @@ Styles:
 * [Column Pinning](column_pinning.md)
 * [Column Resizing](column_resizing.md)
 * [Selection](selection.md)
+@@if (igxName !== 'IgxHierarchicalGrid') {
```
tachojelev: image

The blank lines caused by the **if** statements are reflected within a list.
In order to avoid this, we should probably use a one-liner for the **if** and its body.
**For example:** ```@@if (igxName !== 'IgxHierarchicalGrid') {* [Searching](search.md)}```

PR #1124, opened 2019-03-11
docs(*): update mention of other components in combo and expansion panel

ViktorSlavov:
In a component's topic, when mentioning another component, there should be a link that leads to the mentioned component's topic.


yoshimis:
no further localization required for this PR.

PR #1141, opened 2019-03-19
Add Tree Grid Remote Filtering docs

tachojelev:
IgniteUI/igniteui-angular-samples#1062


jp/components/grids_templates/virtualization.md
```diff
@@ -124,6 +124,82 @@ public processData() {
...
+    });
+}
+```
+
+The remote filtering will have to be performed over the flat collection directly. We will also include all the parents for any record that matches the filtering condition regardless of whether or not the parents match the filtering (we do this to keep the hierarchy intact).
+
+```typescript
+// remoteFilteringService.ts
+
+private filterData(data: any[], filteringArgs: IFilteringExpressionsTree): any[] {
```
DiyanDimitrov: This is a mock-up code which in real life scenario will be performed on the server so it is better to remove this snippet from the docs. We should just mention that when filtering on the server you should include the parents of the matching rows.

PR #1166, opened 2019-04-09
[DO NOT MERGE] Update list topic with display density

tachojelev:
IgniteUI/igniteui-angular#4393

### DIFF
--- a/en/components/list.md
+++ b/en/components/list.md
@@ -321,6 +321,47 @@ toggleFavorite(item: IgxListItem) {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


rmkrmk:
An "of" is missing in the following sentence:
"...we will update our own density property that is bound to the displayDensity **of** the list".

tachojelev:
> An "of" is missing in the following sentence:
> "...we will update our own density property that is bound to the displayDensity **of** the list".

@rmkrmk Issue was addressed.

PR #1170, opened 2019-04-10
docs(drop-down): add virtual drop-down topic

ViktorSlavov:
Related https://github.com/IgniteUI/igniteui-angular-samples/pull/1102

### DIFF
--- a/en/components/drop_down_virtual.md
+++ b/en/components/drop_down_virtual.md
@@ -8,7 +8,7 @@ _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI 
 
 <p class="highlight">[The Ignite UI for Angular Drop Down](drop_down.md) can fully integrate usage of the [IgxForOf directive](for_of.md) in order to display a very large list of items for its selection</p>
 
-<div class="sample-container loading" style="height:220px">
+<div class="sample-container loading" style="height:320px">
     <iframe id="dropdown-virtual-iframe" src='{environment:demosBaseUrl}/data-entries/dropdown-virtual' width="100%" height="100%" seamless="" frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
 </div>
 <div>
@@ -81,9 +81,9 @@ Inside of our component's `constructor`, we'll declare a moderately large list o
 export class DropDownVirtualComponent {
   public items: DataItem[];
   public itemHeight = 48;
-  public itemsMaxHeight = 480;
+  public itemsMaxHeight = 320;
 
-  constructor(protected cdr: ChangeDetectorRef) {
+  constructor() {
     const itemsCollection: DataItem[] = [];
     for (let i = 0; i < 50; i++) {
         const series = (i * 10).toString();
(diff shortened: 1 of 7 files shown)


en/components/drop_down_virtual.md
```diff
@@ -0,0 +1,160 @@
...
+    <iframe id="dropdown-virtual-iframe" src='{environment:demosBaseUrl}/data-entries/dropdown-virtual' width="100%" height="100%" seamless="" frameBorder="0" onload="onSampleIframeContentLoaded(this);"></iframe>
+</div>
+<div>
+<button data-localize="stackblitz" disabled class="stackblitz-btn" data-iframe-id="dropdown-virtual-iframe" data-demos-base-url="{environment:demosBaseUrl}">view on stackblitz</button>
+</div>
+<div class="divider--half"></div>
+
+## Configuration
+
+In order for the drop down to display a list of virtual items, some prerequisites need to be fulfilled.
```
bazal4o: Let's reword a little bit to -> In order to configure the drop down to display a list of virtual items, you need to fulfil some prerequisites

en/components/drop_down_virtual.md
```diff
@@ -0,0 +1,160 @@
...
+> For the drop down to work with a virtualized list of items, [`value`]({environment:angularApiUrl}/classes/igxdropdownitemcomponent.html#value) and [`index`]({environment:angularApiUrl}/classes/igxdropdownitemcomponent.html#index) inputs **must** be passed to all items.
+> [!NOTE]
+> It is strongly advised for each item to have an unique value passed to the `[value]` input. Otherwise, it might lead to unexpected results (incorrect selection). 
+> [!NOTE]
+> When the drop down uses virtualized items, the type of [`dropdown.selectedItem`]({environment:angularApiUrl}/classes/igxdropdowncomponent.html#selecteditem) becomes `{ value: any, index: number }`, where `value` is a reference to the data item passed inside of the `[value]` input and `index` is the item's index in the data set
+
+
+### Component Definition
+
+Inside of our component's `constructor`, we'll declare a moderately large list of items (container both headers and disabled items) which we'll be displaying in our drop-down. We also need to declare `itemHeight` and `itemsMaxHeight` (used in the template) inside of our `drop-down-virtual.component.ts` file:
```
bazal4o: i guess containing instead of container ?

en/components/drop_down_virtual.md
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+    this.items = itemsCollection;
+  }
+}
+```
+
+### Styles
+
+The last (but very important) part of the configuration happens inside of our component's style sheet, `drop-down-virtual.component.scss`. We need to declare that our wrapping div (`drop-down-virtual-wrapper`) has `overflow: hidden`, so that we don't end up with two scroll bars (one from the `igxFor` and one from the container itself):
```
bazal4o: just a suggestion: "we need do declare" -> It's required to set the wrapping div ..... with overflow: hidden. This extra configuration prevents two scrollbars appearance (one from the .....)

ViktorSlavov: I've reworded it to be a bit clearer, changes are in the latest commit

en/components/drop_down_virtual.md
ViktorSlavov: Thanks for pointing it out! Addressed in the latest commit

en/components/drop_down_virtual.md
ViktorSlavov: Addressed in the latest commit

PR #1198, opened 2019-04-16
Update AutoPositioningStartegy

wnvko:
`AutoPositioningStrategy` will now push the shown content if element is out of view even after it was flipped.


kr/components/overlay_position.md
```diff
@@ -22,16 +22,13 @@ Position strategies determine where to display the component in the provided Igx
...
 <div class="divider"></div>
 
-3. **Auto** - Positions the element as in **Connected** positioning strategy and re-positions the element in the view port (calculating a different start point) in case the element is partially out of view. Defaults to:
+3. **Auto** - Positions the element as in **Connected** positioning strategy and re-positions the element in the view port (calculating a different start point) in case the element is partially out of view. **Auto** strategy will try initially to show the element as **Connected** does. If element goes out of the view **Auto** will flip the start point and direction, e.g. if direction is bottom it will switch it to top and so on. If after flipping the element is still out of view **Auto** will use initial directions and start point and will push the element into view, e.g. if the element goes out with 50px in right direction **Auto** will push it with 50px to the left. If after this element is partially out of view, element is wider or higher than the view port, **Auto** will align element's left/top edge with view port's left/top edge. Defaults to:
```
ViktorSlavov: The explanation is clear and concise

A bit of rewording, if you don't mind:
```suggestion
3. **Auto** - Positions the element as in **Connected** positioning strategy and re-positions the element in the view port (calculating a different start point) in case the element is partially out of view. **Auto** strategy will initially try to show the element as **Connected** does. If the element goes out of the view, **Auto** will flip the start point and direction, e.g. if direction is bottom it will switch it to top and so on. After flipping, if the element is still out of view, **Auto** will use the initial direction and start point and will push the element into view, e.g. if the element goes out with 50px in the right direction, **Auto** will push it with 50px to the left. Finally, if the element is partially out of view or if the element is wider or higher than the view port, **Auto** will align the element's left/top edge with the view port's left/top edge. Defaults to:
```

kr/components/overlay_position.md
```diff
@@ -22,16 +22,12 @@ Position strategies determine where to display the component in the provided Igx
...
-
+3. **Auto** - **Auto** - Positions the element the same way as the **Connected** positioning strategy. It also calculates a different starting point in case the element goes partially out of the view port. The **Auto** strategy will initially try to show the element as the **Connected** strategy does. If the element goes out of the view port **Auto** will flip the starting point and the direction, i.e. if the direction is 'bottom', it will switch it to 'top' and so on. After the flipping, if the element is still of the view port, **Auto** will use the initial directions and the starting point to push the element into the view port. For example - if the element goes out of the right side of the view port by 50px, **Auto** will push it with 50px to the left. Afterwards if the element is partially out of the view port, then its height or width were greater than the view port's, **Auto** will align the element's left/top edge with the view port's left/top edge. Defaults to:
```
jackofdiamond5: I think it sounds pretty good, it's easy to read and understand.
A little typo here that you could change
>After the flipping, if the element is still of the view port... 

to

>After the flipping, if the element is still **out** of the view port...

wnvko: Thank you for all the comments. Will apply this now.

yoshimis:
@wnvko Should we make same changes to EN and JP? For JP I will take care of it.

wnvko:
> @wnvko Should we make same changes to EN and JP? For JP I will take care of it.

@yoshimis yes this should be localized. The EN version is already merged. JP and KR are pending.

wnvko:
@yoshimis  Looking again at this there is a mistake - we have changed the KR version which is not localized yet. I will create update now the EN version and you may take care of JP one.

yoshimis:
JP updated: https://github.com/IgniteUI/igniteui-docfx/commit/5e44b1835ebf7fc5f7341491eaca006eebee7dfa

PR #1210, opened 2019-04-23
Update Excel style filtering samples container height

tachojelev:
IgniteUI/igniteui-angular#4367


yoshimis:
no localization necessary for this PR.

PR #1213, opened 2019-04-23
Changes from AngularDocFX_JP

ESShared:


zdrawku:
I see a lot of changes here, my suggestion is to perform sanity testing again on staging after we merge them

Issue #1192, opened 2019-04-15
Angular topics improvements


zdrawku:
Review:
- meta title, description, and keywords
- H1 per page
- topic content requirements
- Images and hyperlinks
- internal links

Topics distribution
**Zdravko:**

- [x] Editing
- [x] Batch Editing
- [x] Filtering
- [x] Sorting
- [x] Summaries
- [x] Column Pinning
- [x] Column Hiding
- [x] Search
- [x] Export excel
- [x] Multi-column headers
- [x] Conditional cell styling

**Hristo:**

- [x] Main grid topic
- [x] Row editing
- [x] Paging
- [x] Excel style filtering
- [x] Group by
- [x] Column moving
- [x] Column resizing
- [x] Selection
- [x] Toolbar
- [x] Paste excel
- [x] Display density

The excel file based on which the changes will be applied is on our SEO internal group.

Remaining topics:

- [ ] Avatar
- [ ] Badge
- [ ] Calendar
- [ ] Carousel
- [ ] Progress bar
- [ ] Dialog
- [ ] Layout manager
- [ ] Slider

zdrawku:
Done for all Grids

PR #1235, opened 2019-05-13
Grid topic enhancements

zdrawku:
Related to #1192


zdrawku:
@hanastasov If you refer to the `meta keywords` there is no difference for the crawler. Same applies for meta title and description. I've decided for better legibility, to use combination of both upper and lower case for the title and description.

hanastasov:
> @hanastasov If you refer to the `meta keywords` there is no difference for the crawler. Same applies for meta title and description. I've decided for better legibility, to use combination of both upper and lower case for the title and description.

I think the meta description is what is seen on the search engine results, so based on this we would want proper formatting, as Angular should be upper case and Material too.. For example searching for "Angular Virtualization directive" directive would bring us:
image
and this is meta description content.

zdrawku:
This is what I've said about the title and description, there I've decided to apply formatting. I see that some descriptions are not covering entirely this, so I am going to update them.

PR #1239, opened 2019-05-14
Update Row Drag topic in JP

ViktorSlavov:
The initial PR looks to be localized, but the translation was done in the wrong file (`jp/components/row_drag.md` instead of `jp/components/grids_templates/row_drag.md`).
The result was the once running `npm run start -- --lang jp`, the `row_drag` page was not properly translated.

This PR moves the already translated text to the proper file.


Lipata:
Verified that samples and Stackblitz are running. Just approve is needed for the JP content.

PR #1252, opened 2019-05-22
Updating the docs for the list component

desig9stein:


simeonoff:
@desig9stein Can you include a theming sample as well.

PR #1268, opened 2019-06-03
Add row reordering sample

jackofdiamond5:


ViktorSlavov:
@jackofdiamond5 @Lipata 
If we're adding a row reordering sample for all 3 of the grids, you can add all of the **docfx** parts in this PR, otherwise you'll have conflicts between the individual additions

en/components/grids_templates/row_drag.md
```diff
@@ -323,6 +333,17 @@ The result of the configuration can be seem below:
 <button data-localize="stackblitz" disabled class="stackblitz-btn" data-iframe-id="hierarchical-grid-row-drag-sample-iframe" data-demos-base-url="{environment:demosBaseUrl}">view on stackblitz</button>
 </div>
 <div class="divider--half"></div>
+
+The following sample demonstrates row reordering in the hierarchical grid.
```
ViktorSlavov: I think this would be better as:
```suggestion
The following sample demonstrates how to configure row reordering in the hierarchical grid using row dragging.
```

ViktorSlavov: If you think the above wording words, please also apply it to the other 2 grid paragraphs (grid and tree-grid)

PR #1289, opened 2019-06-12
docs(*): Updated time and date picker topics #4647

sboykova:
Related to IgniteUI/igniteui-angular#4647


bkulov:
Related samples PR IgniteUI/igniteui-angular-samples#1196

PR #1216, opened 2019-05-07
Multi row layout topic

dkamburov:

PR #1315, opened 2019-06-25
docs(grid): Added styling section in the resizing topics for grids

sboykova:
Related to  #1216

### DIFF
--- a/en/components/grids_templates/column_resizing.md
+++ b/en/components/grids_templates/column_resizing.md
@@ -20,7 +20,6 @@ _keywords: Ignite UI for Angular, UI controls, Angular widgets, web widgets, UI 
 ---
 }
 
-
 ### @@igComponent Column Resizing
 
 With deferred column resizing, the user will see a temporary resize indicator while the drag operation is in effect. The new column size is only put into its place once the drag operation has ended.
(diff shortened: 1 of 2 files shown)


en/components/grids_templates/column_resizing.md
```diff
@@ -167,7 +166,126 @@ column.autosize();
...
+@import '~igniteui-angular/lib/core/styles/themes/index';
+```
+
+The simplest approach to achieve this is to create a new theme that extends the [`igx-grid-theme`]({environment:sassApiUrl}/index.html#function-igx-grid-theme) and accepts many parameters as well as the `$resize-line-color` parameter.
+
+``` scss
+$blue-color: #0288D1;
+$custom-grid-theme: igx-grid-theme(
+    $resize-line-color: $blue-color
+);
```
HristoP96: The first demonstration of the theme creation should be the the most hard-coded, without CSS variables.
`$custom-grid-theme: igx-grid-theme(
    $resize-line-color:  #0288D1
);`
This should be enough.

en/components/grids_templates/column_resizing.md
```diff
@@ -167,7 +166,126 @@ column.autosize();
...
+));
+
+// Specifying the palette and schema of the custom grid theme
+$custom-grid-theme: igx-grid-theme(
+    $palette: $custom-theme-palette,
+    $schema: $custom-light-grid-schema
+);
+```
+Don't forget to include the theme in the same way as it was demonstrated above.
+
```
HristoP96: It will be better if there is a `####Demo` section, so that the reader could have faster reference to the sample.

PR #1346, opened 2019-07-04
Add Multi-Cell Selection Styling sections

ViktorSlavov:


en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+Once done, we can make use of the [`igx-contrast-color`]({environment:sassApiUrl}/index.html#function-igx-contrast-color) and [`igx-color`]({environment:sassApiUrl}/index.html#function-igx-color) functions. With them, we define the colors we would like to use for our selection range:
+```scss
+    $text-color: igx-contrast-color($default-palette, 'primary', 900);
+    $background-color: igx-color($default-palette, "primary", 900);
+    $border-yellow: #f2c43c;
+```
+
+#### Create custom theme
+ 
+Next we create need to create a new theme that extends the [`igx-grid-theme`]({environment:sassApiUrl}/index.html#function-igx-grid-theme) passing our `text-color`, `background-color` and `border-yellow` variables as `$cell-selected-text-color`, `$cell-selected-background` and `$cell-active-border-color`, respectively:
```
damyanpetev: "Next we **create** need to **create**"

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+
+```scss
+$custom-grid-theme: igx-grid-paginator-theme(
```
damyanpetev: paginator theme?

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+    $cell-selected-text-color: $text-color,
+    $cell-active-border-color: $border-yellow,
+    $cell-selected-background: $background-color
+);
+```
+
+#### Consume theme
+
+Afterwards, all we need to do is have our @@igSelector consume the newly created theme instead of the default one:
```
damyanpetev: The grid (`@@igSelector`) does the consuming of sorts, but what you are doing here is [including the theme mixin](https://sass-lang.com/documentation/at-rules/mixin) in the [app] component styles to override the default theme for the grid inside.

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+```scss
+    :host { 
+        ::ng-deep {
+            @include igx-grid($custom-grid-theme);
+        }
+    }
+```
+
+ >[!NOTE]
+ >If the component is using an [`Emulated`](../themes/component-themes.md#view-encapsulation) ViewEncapsulation, it is necessary to `penetrate` this encapsulation using `::ng-deep`:
```
damyanpetev: Make the whole phrase "Emulated ViewEncapsulation" the link. Also keep the back tick (`` ` ``) for code-related references.

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+
+We scope the style under `:host` selector so as not to affect any other grids we might have in our application.
```
damyanpetev:
```suggestion
We scope the style under `:host` selector so as to not affect any other grids we might have in our application.
```

damyanpetev: Ugh, this might require a bit more explanation as well.
The `:host` pseudo-selector will only work in Emulated encapsulation or in native shadow DOM. If you go with `None` encapsulation (which I'm sure some do) the host will do exactly.. nothing.
⚠️ I.e. this explanation and the snippet above should clarify they apply only when using encapsulation.

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,99 @@ public selection = true;
...
+
+As a result of the styling, the selected grid cells are highlighted with a more distinctive color:
```
damyanpetev:
```suggestion
With the custom theme applied, the selected grid cells are highlighted with our selected colors:
```
Or something like this...

en/components/grids_templates/selection.md
damyanpetev: Applies to the title as well. "Apply" would be more suitable.

en/components/grids_templates/selection.md
```diff
@@ -363,34 +363,39 @@ Once done, we can make use of the [`igx-contrast-color`]({environment:sassApiUrl
...
     :host { 
         ::ng-deep {
             @include igx-grid($custom-grid-theme);
         }
     }
 ```
-
  >[!NOTE]
- >If the component is using an [`Emulated`](../themes/component-themes.md#view-encapsulation) ViewEncapsulation, it is necessary to `penetrate` this encapsulation using `::ng-deep`:
+ > With `Emulated` or `ShadowDОМ` view encapsulation, we scope the style under `:host` selector so as not to affect any other grids we might have in our application.
```
damyanpetev: Okay, not a fan of the two notes.
`:host` scoping is only really needed for `Emulated` and thus should probably be explained in the note above with it.
Basically, `::ng-deep` disables the emulated encapsulation (adding attrs) for its nested styles and Shadow DOM natively does the scoping already.

en/components/grids_templates/selection.md
```diff
@@ -339,6 +339,104 @@ public selection = true;
...
+$custom-grid-theme: igx-grid-theme(
+    $cell-selected-text-color: $text-color,
+    $cell-active-border-color: $border-yellow,
+    $cell-selected-background: $background-color
+);
+```
+
+#### Apply theme
+
+Afterwards, all we need to do is include the mixin in our component's style (this could also be done in the `app.root`), so that our @@igSelector uses the newly created theme instead of the default one:
```
damyanpetev:
```suggestion
Afterwards, all we need to do is include the mixin in our component's style (could also be in the app styles), so that our @@igSelector uses the newly created theme instead of the default one:
```

PR #1381, opened 2019-07-17
Adds docs for GroupBy w/ Paging

3phase:


mpavlinov:
Removing "status: pending-localization", because there is a newer PR.

PR #1384, opened 2019-07-17
Add banner styling topic section

IvayloG:


en/components/banner.md
```diff
@@ -234,6 +234,79 @@ export class MyBannerComponent {
...
+);
+```
+
+#### Applying
+All that's left is to properly scope our newly created theme.
+
+##### Globally
+In case you want this newly created `igx-banner` theme to be applied [`globally`](themes/component-themes.md#creating-themes) in your app, all that is needed is to include the theme in your app root style file:
+```scss
+// in root app.scss
```
ViktorSlavov: This is in `styles.scss`, `app.scss` will apply only to the elements defined in `app.component.ts`

IvayloG: Good catch

en/components/banner.md
```diff
@@ -234,6 +234,101 @@ export class MyBannerComponent {
 > [!NOTE]
 > If the changes above are applied, the banner will never open, as the opening event is always cancelled.
 
+### Styling
```
ViktorSlavov: I think the styling section should h2 (`##` instead of `###`), so it appears on the right navigation right away, since it's a major section by itself:
image

PR #1366, opened 2019-07-12
Adding Tabs/BottomNav Routing Docs

rmkrmk:
IgniteUI/igniteui-angular#4297

PR #1385, opened 2019-07-17
Adding routing for tabs/bottomNav in JP files

rmkrmk:
IgniteUI/igniteui-angular#4297


DiyanDimitrov:
@yoshimis, We forgot to add the JP files in the original PR #1366 and we created this PR. Should we merge this one now or should we wait for the JP localization?

PR #1389, opened 2019-07-18
Add checkbox styling section

IvayloG:
Related https://github.com/IgniteUI/igniteui-angular-samples/pull/1307

### DIFF
--- a/en/components/checkbox.md
+++ b/en/components/checkbox.md
@@ -99,7 +99,7 @@ First, in order for us to use the functions exposed by the theme engine, we need
 
 #### Define palette
 We are going to use our own color [palette](themes/palette.md) where we can specify our two main theme colors. Additionally we are going to set several of the exposed by the [`igx-checkbox-theme`]({environment:sassApiUrl}/index.html#function-igx-checkbox-theme) parameters namely `$border-radius`, `$label-color` and `$empty-color`.
-Fist define a custom palette:
+First define a custom palette:
 ```scss
 // in styles.scss
 $my-primary-color: #f5e492;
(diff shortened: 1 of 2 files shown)


en/components/checkbox.md
```diff
@@ -86,6 +86,81 @@ The final result would be something like that:
...
+First, in order for us to use the functions exposed by the theme engine, we need to import the `index` file in our style file: 
+
+```scss
+// in styles.scss
+@import '~igniteui-angular/lib/core/styles/themes/index';
+```
+
+#### Define palette
+We are going to use our own color [palette](themes/palette.md) where we can specify our two main theme colors. Additionally we are going to set several of the exposed by the [`igx-checkbox-theme`]({environment:sassApiUrl}/index.html#function-igx-checkbox-theme) parameters namely `$border-radius`, `$label-color` and `$empty-color`.
+Fist define a custom palette:
```
ViktorSlavov: `First`

en/components/checkbox.md
```diff
@@ -86,6 +86,81 @@ The final result would be something like that:
...
+$my-secondary-color: #09f;
+
+$my-color-palette: igx-palette(
+    $primary: $my-primary-color,
+    $secondary: $my-secondary-color
+);
+```
+
+In order to see our custom palette applied, we need to pass it to a theme function.
+So in one bold move we will [`create a custom theme`](themes/component-themes.md#creating-themes) and pass our tree more specific parameters as well. Let's say we have decided modifying these will be more than sufficient to make our component look the way we like. It is really convenient to take use of the previously created [palette](themes/palette.md) and base our new colors on the colors defined.
```
ViktorSlavov: We're not exactly making use of the palette here, so we can drop the last part
```suggestion
So in one bold move we will [`create a custom theme`](themes/component-themes.md#creating-themes) and pass our three specific parameters as well. Let's say we have decided modifying these will be more than sufficient to make our component look the way we like:
```

en/components/checkbox.md
```diff
@@ -86,6 +86,81 @@ The final result would be something like that:
...
+##### Globally
+In case you want this newly created `igx-checkbox` theme to be applied [`globally`](themes/component-themes.md#creating-themes) in your app, all that is needed is to include the theme in your app root style file:
+```scss
+// in styles.scss
+// Pass our checkbox theme to the `igx-checkbox` mixin
+    @include igx-checkbox($custom-checkbox-theme);
+```
+
+##### Scoped
+There may be a case where you want a particular `igx-checkbox` be styled differently than the others in the app. This will require to use angular specific pseudo-class selectors like `:host`, `::ng-deep`, etc. Additionally all of the above steps need to be moved from styles.scss to the component.scss file.
```
ViktorSlavov:
```suggestion
There may be a case where you want a particular `igx-checkbox` be styled differently than the others in the app. This will require to use Angular specific pseudo-class selectors like `:host`, `::ng-deep`, etc. Additionally, all of the above styles we specify in styles.scss need to be defined in the component.scss file instead.
```

PR #1396, opened 2019-07-19
Adds docs for igxForOf's new local variables

3phase:


mpavlinov:
Removing "status: pending-localization", because there is a newer PR.

PR #1419, opened 2019-07-25
Document igxForOf known issue when working with hidden containers like igxTabs

### LINKED ISSUE #1270: Document igxForOf/igxGrid etc. known issue when working with hidden containers like igxTabs
There is a problem when igxForOf (and as a consequence igx*Grid*) is in a hidden containers.
The issues is logged in igniteui-angular project, but since we cannot fix it we will document it: https://github.com/IgniteUI/igniteui-angular/issues/4884

3phase:
Closes #1270 
For reference: https://github.com/IgniteUI/igniteui-angular/issues/4884


en/components/grid/grid.md
```diff
@@ -616,6 +616,8 @@ public ngOnInit() {
...
+| Scroll position of components that use igxForOf is not preserved |When changing the context of an element, that uses igxForOf, the scrollbar position would not be preserved, even though the data position would be. That inconsistency occurs when the content is created and after that the DOM nodes get changed. In such a case, the nodes are getting detached from the DOM and then reattached again in a different place. That's why they will not keep their scroll state, but ever since the component is not destroyed/recreated, it will still have its internal state. Some possible workarounds are: <br/> <ul><li>Resetting the DOM node's state, for example by wrapping it in an ngIf, that'd follow it's state (the tab gets selected => attaching nodes to the DOM tree, the tab gets disselected => removing nodes from the tree).</li><li>Persisting the state, i.e. determining whether the element has been re-added to the DOM tree and resetting its scroll position manually.</li></ul>
```
mpavlinov: I think the description should be `igxGrid` specific. Here's what I suggest:
|`igxGrid` won't preserve its scroll position if put in projected container like `<ng-content>`. | This is the case with `igxGrid` in `igxTabs`. When tabs get changed the grid won't preserve its scrollbar position, altough the grid content is at correct offset in the data source. The reason is that `igxGrid` uses `igxForOf` directive internally, which cannot identify the detaching and re-attaching of the DOM elements from the DOM tree and thus cannot update the scrollbar thumb position.|

en/components/for_of.md
```diff
@@ -200,6 +200,12 @@ The following code snippet demonstrates how to use the `even` property in an `ng
 ```
 In the example above, an `even` class is assigned to every even div element.
 
+## Known Limitations
+
+|Limitation|Description|
+|--- |--- |
+| Scroll position of components that use igxForOf is not preserved |When changing the context of an igxForOf-using element, the scrollbar position won't be preserved, even though the data position would be. That inconsistency occurs when the content is created and after that the DOM nodes get changed. In such a case, the nodes are getting detached from the DOM and then reattached again in a different place. That's why they will not keep their scroll state, but ever since the component is not destroyed/recreated, it will still have its internal state. Some possible workarounds are: <br/> <ul><li>Resetting the DOM node's state, for example by wrapping it in an ngIf</li><li>Persisting the state, i.e. determining whether the element has been re-added to the DOM tree and resetting its scroll position manually</li></ul>
```
mpavlinov: Here's what I suggest:
| Scroll position of components that use igxForOf is not preserved in projected container |When component that uses `igxForOf` directive is in `<ng-content>` or other projected container its scrollbar position won't be preserved, even though the data position would be. That's because the DOM elements are detached and later re-attached to the DOM tree as a result losing the scrollbar scroll position.
Some possible workarounds are: <br/> <ul><li>Resetting the DOM node's state, for example by wrapping it in an `ngIf`. As a result the component will be destroyed and then re-created loosing all its state.</li><li>Persisting the state, i.e. determining whether the element has been re-added to the DOM tree and resetting its scroll position manually. For example using [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver).</li></ul>

3phase: Done

en/components/grid/grid.md
3phase: Done

mpavlinov:
Closing this one (because it became too complex) and making a new PR: https://github.com/IgniteUI/igniteui-docfx/pull/1460

PR #1441, opened 2019-08-07
docs(grid): Update cell templating, add CRUD events

### LINKED ISSUE #1352: Grid CRUD documentation - incomplete and misleading
The current section on [editing](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/editing.html) in the grids is both incomplete and misleading.

Main pain points:
- No mention of custom editor templates. There is some information [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/grid.html#columns-configuration) but it tricks the user into binding to the wrong property of the template context.
- No mention of available CRUD events, when they are thrown and sample usage with them.
- What is the difference between providing a default cell template and an editor cell template.

ViktorSlavov:
Closes #1352


en/components/grids_templates/editing.md
```diff
@@ -152,9 +158,23 @@ Another way to update cell is directly through [`update`]({environment:angularAp
...
+```html
+<igx-column field="ReorderLevel" header="Reorder Level" [editable]="true">
+    <ng-template igxCellEditor let-cell="cell">
+        <input type="number" [(ngModel)]="cell.editValue" />
+    </ng-template>
+</igx-column>
+```
+
+> [!NOTE]
+> The cell editing template directive (`igxCellEditor`) is very different from the [cell template(`igxCell`)](../grid/grid.md#cell-template) - the former handles how a column's cells **in edit mode** are displayed an controls the edited cell's **edit value**, the latter control how a column's cell's are show when **outside of edit mode**.
```
ViktorSlavov: @rkaraivanov , could you take a look when you've got the time?
Does this make a clear enough difference between **cell template** and **cell editing template**, or should I add some extra explanation?

rkaraivanov: @ViktorSlavov 

IMO, I think the main distinction should be between the functional differences between the two directives. In the end, they both allow for customizing the presentation of cells. Thus an user can still re-template the default cell content through the `igxCell` directive and achieve CRUD functionality even when the column may not be editable. For example:

```html
<ng-template igxCell let-value>
     <!-- This will update the underlying record even when the column is not editable -->
    <input [(ngModel)]="value" /> 
</ng-template>
```

What he won't get however are all the CRUD events, transaction support, etc. I think that should be the focus point of the documentation, backed up by some code snippets/samples which actually show the difference.

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,162 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+| `onCellEditCancel` | Fires when a cell exits edit mode **without committing** its value (e.g. by pressing `Escape`) |
+| `onRowEdit` | Fires when a row in edit mode's value is **committed** (e.g. by clicking the `Done` button on the Row Editing Overlay) |
+| `onRowEditCancel` | Fires when a row exits edit mode **without committing** its values (e.g. by clicking the `Cancel` button on the Row Editing Overlay) |
+
+All of the above events can be cancelled. For example, if `onCellEditEnter` is cancelled, the cell will never enter edit mode.
+
+#### Editing Events - Configuration
+Using the grid's editing events we can alter how the user interacts with the grid.
+In the below example, we'll validate a cell based on the data entered in it by binding to the [`onCellEdit`]({environment:angularApiUrl}/classes/@@igTypeDoc.html#oncelledit) event. If the new value of the cell does not meet our predefined criteria, we'll display a custom error message (using [`IgxToast`](../toast.md). We'll also cancel the editing event (by setting `event.cancel = true`), so the user has to specify a valid value before continuing with the editing.
```
damyanpetev:
```suggestion
In the below example, we'll validate a cell based on the data entered in it by binding to the [`onCellEdit`]({environment:angularApiUrl}/classes/@@igTypeDoc.html#oncelledit) event. If the new value of the cell does not meet our predefined criteria, we'll display a custom error message using [`IgxToast`](../toast.md). We'll also cancel the editing event (by setting `event.cancel = true`), so the user has to specify a valid value before continuing with the editing.
```

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,162 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+
+The first thing we need to is bind to the grid's event:
+
+```html
+<@@igSelector (onCellEdit)="handleCellEdit($event)"
+...>
+...
+</@@igSelector>
+```
+@@if (igxName === 'IgxTreeGrid') {
```
damyanpetev: _That's Hierarchical Grid config below, no?_

damyanpetev: Also see https://github.com/IgniteUI/igniteui-angular/issues/5611 - this won't be the case for too long, so I think avoiding this section might actually be better. Plus the `columnList` lookup in the snippet after will only work on the root grid anyways.

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,162 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+        }
+    }
+}
+```
+Since we're passing the grid as a second argument in the `handleCellEdit` calls, we're using the passed grid's columns to find to which column the cell belongs. With the above validation in place, when a user tries to change an artist's **Debut** year or an album's **Launch Date**, the grid will not allow any dates that are greater than today.
+}
+
+#### Editing Events - Demo
+
+The result of the above validation being applied to our @@igSelector can be seen in the below demo:
```
damyanpetev:
```suggestion
The result of the above validation being applied to our `@@igSelector` can be seen in the below demo:
```

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,162 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+                event.cancel = true;
+            }
+        }
+    }
+}
+```
+Since we're passing the grid as a second argument in the `handleCellEdit` calls, we're using the passed grid's columns to find to which column the cell belongs. With the above validation in place, when a user tries to change an artist's **Debut** year or an album's **Launch Date**, the grid will not allow any dates that are greater than today.
+}
+
+#### Editing Events - Demo
```
damyanpetev: _You don't need to split this section TBH. The entire section can be called "Events Configuration example" or something_

en/components/grids_templates/editing.md
ViktorSlavov: Correct, it's for the HGrid, thanks for pointing it out!

en/components/grids_templates/editing.md
ViktorSlavov: It's convenient if you want to send a link to the demo - a user won't have to scroll through all of the configuration and can just lookup the `StackBlitz` example, in order to see if the implemented functionality suits their needs

en/components/grids_templates/editing.md
```diff
@@ -152,9 +158,26 @@ Another way to update cell is directly through [`update`]({environment:angularAp
 ```
 }
 
-@@if (igxName === 'IgxGrid') {
-If you want to define a custom template which will be applied when the cell is in edit mode, you can see the documentation for [Grid Columns configuration](grid.md#columns-configuration).
-}
+#### Cell Editing Templates
+
+If you want to provide a custom template which will be applied when a cell is in edit mode, you can make use of the [`igxCellEditor` directive]({environment:angularApiUrl}/classes/igxcelltemplatedirective.html). In order to use a custom cell editing template for a column, you need to pass an `ng-template` with the `igxCellEditor` directive and properly bind your custom control to the [`cell.editValue`]({environment:angularApiUrl}/classes/igxcellcomponent.html#editvalue):
```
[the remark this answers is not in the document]
ViktorSlavov: Not exactly, but yeah, I can merge them into one. The first part is "use this when...", the second part - "this is how you use it".

en/components/grid/grid.md
```diff
@@ -157,13 +170,15 @@ to set the [`editable`]({environment:angularApiUrl}/classes/igxcolumncomponent.h
         <label for="price">
             Enter the new price tag
         </label>
-        <input name="price" type="number" [ngModel]="cell.value" (ngModelChange)="cell.update(convertToNumber($event))" />
+        <input name="price" type="number" [ngModel]="cell.editValue" (ngModelChange)="cell.update(convertToNumber($event))" />
```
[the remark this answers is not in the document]
ViktorSlavov: Thanks, good catch - it's leftover from the previous code snippet, I should remove the `ngModelChange` handling

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,144 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+| [`onCellEditCancel`]({environment:angularApiUrl}/classes/igxgridcomponent.html#oncelleditcancel) | Fires when a cell exits edit mode **without committing** its value (e.g. by pressing `Escape`) | [IGridEditEventArgs]({environment:angularApiUrl}/interfaces/igridediteventargs.html) | `true` |
+| [`onRowEdit`]({environment:angularApiUrl}/classes/igxgridcomponent.html#onrowedit) | Fires when a row in edit mode's value is **committed** (e.g. by clicking the `Done` button on the Row Editing Overlay) | [IGridEditEventArgs]({environment:angularApiUrl}/interfaces/igridediteventargs.html) | `true` |
+| [`onRowEditCancel`]({environment:angularApiUrl}/classes/igxgridcomponent.html#onroweditcancel) | Fires when a row exits edit mode **without committing** its values (e.g. by clicking the `Cancel` button on the Row Editing Overlay) | [IGridEditEventArgs]({environment:angularApiUrl}/interfaces/igridediteventargs.html) | `true` |
+
+All of the above events can be cancelled. For example, if `onCellEditEnter` is cancelled, the cell will never enter edit mode.
+
+#### Cell validation via edit event cancelling
```
damyanpetev:
```suggestion
#### Cell validation on edit event
```
_Not really validation with the cancel, more the other way around. Also the cancelling bit has different significance_

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,144 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+Using the grid's editing events we can alter how the user interacts with the grid.
+In the below example, we'll validate a cell based on the data entered in it by binding to the [`onCellEdit`]({environment:angularApiUrl}/classes/@@igTypeDoc.html#oncelledit) event. If the new value of the cell does not meet our predefined criteria, we'll display a custom error message using [`IgxToast`](../toast.md). We'll also cancel the editing event (by setting `event.cancel = true`), so the user has to specify a valid value before continuing with the editing.
```
damyanpetev: Perhaps:
```suggestion
In this example, we'll validate a cell based on the data entered in it by binding to the [`onCellEdit`]({environment:angularApiUrl}/classes/@@igTypeDoc.html#oncelledit) event. If the new value of the cell does not meet our predefined criteria, we'll prevent it from reaching the data source by cancelling the event (`event.cancel = true`). We'll also display a custom error message using [`IgxToast`](../toast.md). 
```
_Swapped those around as "has to specify a valid value before continuing with the editing" isn't always valid, and the important bit from the cancel is that it doesn't commit the data._

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,144 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+                this.toast.message = "The employee hire date must be in the past!";
+                this.toast.show();
+            }
+        }
+    }
+}
+```
+Here, we are validating two columns. If the user tries to set an invalid value for an employee's **Age** (below 18) or their hire date (a future date), the editing will be cancelled (the value will not be submitted) and a toast with an error message will be displayed.
+}
+@@if (igxName === igxName === 'IgxHierarchicalGrid') {
```
damyanpetev: This snippet doesn't display for me. Maybe the extra equals confuses check?
```suggestion
@@if (igxName === 'IgxHierarchicalGrid') {
```

en/components/grids_templates/editing.md
damyanpetev: The tree grid text uses something like "the editing will be cancelled (the value will not be submitted)" which is also fine

en/components/grids_templates/editing.md
```diff
@@ -288,6 +311,144 @@ These can be wired to user interactions, not necessarily related to the **@@igSe
...
+            if (event.newValue > new Date().getTime()) {
+                event.cancel = true;
+                this.toast.message = "The employee hire date must be in the past!";
+                this.toast.show();
+            }
+        }
+    }
+}
+```
+Here, we are validating two columns. If the user tries to set an invalid value for an employee's **Age** (below 18) or their hire date (a future date), the editing will be cancelled (the value will not be submitted) and a toast with an error message will be displayed.
```
damyanpetev:
```suggestion
Here, we are validating two columns. If the user tries to set an invalid value for an employee's **Age** (below 18) or their **Hire Date** (a future date), the editing will be cancelled (the value will not be submitted) and a toast with an error message will be displayed.
```

ViktorSlavov:
Newly added sections in the **JP** docs are *commented out* - please uncomment them, once they're translated.

PR #1450, opened 2019-08-13
Update transaction topic according to the new demo

Lipata:


en/components/transaction-how-to-use.md
```diff
@@ -23,197 +23,280 @@ In this topic we will use [`igxList`]({environment:angularApiUrl}/classes/igxlis
...
-        case TransactionType.ADD:
-            return "green";
-        case TransactionType.DELETE:
-            return "red";
-        case TransactionType.UPDATE:
-            return "blue";
-        default:
-            return null;
-    }
+Get transactions, merge with wishlist data for display purposes
```
ViktorSlavov: Not sure if this line should be here, seems misplaced

PR #1460, opened 2019-08-15
Document igxForOf/igxGrid etc. known issue when working with hidden containers like igxTabs

### LINKED ISSUE #1270: Document igxForOf/igxGrid etc. known issue when working with hidden containers like igxTabs

mpavlinov:
Closes #1270
For reference: IgniteUI/igniteui-angular#4884


yoshimis:
Already localized in other PRs.

Issue #1466, opened 2019-08-20
Incomplete sentence in sizing.md


yoshimis:
Incomplete sentence in https://staging.infragistics.com/products/ignite-ui-angular/angular/components/grid/sizing.html.

1. Look for the text below:
In the next example the columns have width of 300px and grid has width if 1200px, which means that the columns have excess of ???

Result:
Some text is missing after "excess of" in the end.

PR #1468, opened 2019-08-20
Adding new topics.

yoshimis:


yoshimis:
In sizing.md, there are some EN left: "In the next example the columns have width of `300px` and grid has width if `1200px`, which means that the columns have excess of". I will translate once #1466 is fixed.

yoshimis:
@valadzhov fixed.

Issue #1275, opened 2019-06-06
Add styling section to all features that can be customized


zdrawku:
Start with the Grid topics below, in that order. Feel free to change the concept of styling if needed:

**Phoenix**
--
**GRIDS**
- [x] Summaries - Demonstrate the summary-theme
- [x] Grid Paging topic may show how to change the default styling of the paginator
- [ ] Main Grid topic - demonstrate row-odd-background, row-even-background, row-odd-text-color, row-hover-background, row-border-color, row-highlight etc.
- [x] Sorting - demonstrate how to change the default icon and color of the sort direction (.igx-grid__th-icons .sort-icon)
- [x] Multi column headers - change the header colors based on parent group
- [ ] Grid crud
- [ ] Column component
- [ ] Cell component

-- 
**NON GRIDS**
- [x]  - Avatar
- [x] - Badge
- [x] - Calendar
- [x] - Carousel
- [x] - Circular Progress bar
- [x] - Dialog
- [x] - Linear Progress bar
- [x] - Slider
........................................................................................................

**Grinders**
--
**GRIDS**
- [x] GroupBy - Use all group parameters, to demonstrate different color variations, grouparea-background, group-row-background, group-row-selected-background, group-label-column-name-text, group-label-icon and group-label-text, group-count-background, group-count-text-color. Also, show how to use the template in order to hide or change the default icon or change the Grouped content entirely.
- [x] Column pinning - Change the pinned-border-width, border-style and border color. 
- [x] Multi Row Layout - Demonstrate how to style the multi row layout cells.
- [x] Hierarchical Grid main topic
- [x] Pinning

-- 
**NON GRIDS**
- [x] - Chip
- [x] - Icon
- [x] - Ripple
........................................................................................................

**Samsara**
--
**GRIDS**
- [x] Column moving - style the draggable header - ghost-header-background, ghost-header-text-color and icon color;
- [x] Column resizing - Demonstrate usage of resize-line-color
........................................................................................................

**CodeX**
--
**GRIDS**
- [x] Filtering  - demonstrate changing of filtering-header-background and filtering-row-background.
- [x] Excel style filtering - think of a way to demonstrate filter dialog styling and typography, consider toggle-filtered-background and toggle-icon-filtered-color as well
- [x] Column hiding - demonstrate igx-column-hiding map
- [x] Toolbar - Demonstrate the igx-grid-toolbar theme
- [x] Tree Grid main topic

-- 
**NON GRIDS**
[...]

PR #1469, opened 2019-08-20
Adds Column Pinning Styling Topic

3phase:
Related to #1275


VMihalkov:
Looks like a text is missing:
image

I think that the apostrophe should be removed:
image

3phase:
> Looks like a text is missing:
> image
> 
> I think that the apostrophe should be removed:
> image

Done

yoshimis:
Localized in #1517 .

PR #1421, opened 2019-07-25
Add How to use required section 

IvayloG:
closes #1417

PR #1475, opened 2019-08-21
Add declarative item and label.

IvayloG:
Related https://github.com/IgniteUI/igniteui-docfx/pull/1421


yoshimis:
Updated in #1518.

PR #1477, opened 2019-08-21
Adds Hierarchical Grid Styling Topic

3phase:
Related to #1275


VMihalkov:
Why are we talking about styling of "group by" in the topic for hierarchical grid:
image

Text is missing:
image

There is no "Demo" subtitle above the sample.

3phase:
> Why are we talking about styling of "group by" in the topic for hierarchical grid:
> image
> 
> Text is missing:
> image
> 
> There is no "Demo" subtitle above the sample.

Done

yoshimis:
Updated in #1532.

PR #1512, opened 2019-09-03
docs(general): SEO optimizations for WPFtoAngular #1447

astaev:
#1447


zdrawku:
The JP changes should be made to vNext, @DiyanDimitrov is familiar with that

PR #1559, opened 2019-10-01
Update select groups section

### LINKED ISSUE #1481: Update select groups section
### Description  
Related to https://github.com/IgniteUI/igniteui-angular-samples/issues/1419 as there are snippets and explanation that needs to update to match the demo code.

While at it, this section is rather confusing:
"You would notice that now we pass in objects that have certain properties, such as type and fruits. This is because the IgxSelectItemComponent has functionality that allows it to receive specific styling inside the drop-down list. This functionality comes inherited from the IgxDropDownItemComponent."

If this is referring to the groping item - that's a separate component so not sure what the passing and styling is about.

IvayloG:
Closes #1481

Related https://github.com/IgniteUI/igniteui-angular-samples/pull/1501

### DIFF
--- a/en/components/select.md
+++ b/en/components/select.md
@@ -268,30 +268,21 @@ Another thing worth mentioning is that `igx-select` uses the `SelectPositioningS
 
 <button data-localize="stackblitz" disabled class="stackblitz-btn" data-iframe-id="select-sample-2-iframe" data-demos-base-url="{environment:demosBaseUrl}">view on stackblitz</button>
 
-Thanks to the fact that `igx-select` extends `igx-drop-down` it also has built-in support for *groups*. 
-
-In order to make use of this functionality you need to change the data that will be passed to `igx-select`, which in this case should look something like this:
+Select component exposes items grouping functionality. This helps to visually group and indent different items based on some criteria.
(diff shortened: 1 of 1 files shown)


en/components/select.md
```diff
@@ -268,30 +268,21 @@ Another thing worth mentioning is that `igx-select` uses the `SelectPositioningS
 
 <button data-localize="stackblitz" disabled class="stackblitz-btn" data-iframe-id="select-sample-2-iframe" data-demos-base-url="{environment:demosBaseUrl}">view on stackblitz</button>
 
-Thanks to the fact that `igx-select` extends `igx-drop-down` it also has built-in support for *groups*. 
-
-In order to make use of this functionality you need to change the data that will be passed to `igx-select`, which in this case should look something like this:
+Select component exposes items grouping functionality. This helps to visually group and indent different items based on some criteria.
+In order to make use of this feature, you may change the data to contain a grouping criteria. In the following case, we use the `label` for the purpose and list a collection of `items` under each group:
```
damyanpetev: Change the data? I don't see anything similar described before TBH.
```suggestion
This works best with hierarchical data that can be iterated declare the components. In the following example, each group has a `label` and list a collection of `items`:
```

PR #1592, opened 2019-10-14
RAndriova/0927

randriova:


yoshimis:
@zdrawku We would like to deploy 8.2 changes in staging to production soon. In that case, should this PR be merged to master instead of vNext then?

zdrawku:
@yoshimis, in either case, all changes should be PRed to master as well. For the JP version, we are performing manual deploy from `master`, which means that if vNext (staging) contains the exact same changes as master, once we deploy master - staging will be the same as production.
The difference from the old process is that we now have two different definitions for staging and production. Back in the days, we were promoting just the approved version of Staging to Production, now they are separate builds.

zdrawku:
@valadzhov and @randriova if you feel any uncertainties in cherry-picking the changs to master, let me know

PR #1596, opened 2019-10-15
Updating JP according to UI review in context.

yoshimis:


yoshimis:
@randriova There is no corresponding JP page for Angular Introduction to Components in the original Angular site. I have added (英語) at the end of title.

PR #1649, opened 2019-11-07
Adding master-detail topic.

MayaKirova:


dkamburov:
We should probably extend the sample or reduce the iframe of the sample
image
Also the **View in Stackblitz** button is not configured

PR #1665, opened 2019-11-18
Changes from AngularDocFX_EN

ESShared:


en/components/data-chart-navigation.md
```diff
@@ -85,7 +85,7 @@ The following code snippet demonstrates how you can enable UI navigation in the
 
 <!-- ### Chart Navigation with Overview Plus Detail Pane
 
-In the `Ig$DataChart` control, there is an overlaid control that allows navigation. This control supports a preview of most supported series types as well as all navigation methods described above. This overlay is the overview plus detail pane, and it can be enabled by setting the `overviewPlusDetailPaneVisibility` property.
+In the `XamDataChart` control, there is an overlaid control that allows navigation. This control supports a preview of most supported series types as well as all navigation methods described above. This overlay is the overview plus detail pane, and it can be enabled by setting the `overviewPlusDetailPaneVisibility` property.
```
kdinev: Should this be `XamDataChart`?

en/components/data-chart-legends.md
```diff
@@ -2,7 +2,7 @@
 title: Data Chart | Visualization Tools | Ignite UI for Angular | Infragistics | Legends
 _description: Create a data chart that displays multiple instances of visual elements in the same plot area in order to create composite chart views.
 _keywords: data chart, Ignite UI for Angular, Infragistics
-mentionedTypes: ['XamDataChart', 'Infragistics.Controls.Charts.Legend']
+mentionedTypes: ['XamDataChart', 'XamLegend','Infragistics.Controls.Charts.Legend']
```
kdinev: Should this be `XamDataChart`, `XamLegend`?

en/components/data-chart-navigation.md
agoldenbaum: Yes, the translations are done from the C# type.

en/components/data-chart-legends.md
agoldenbaum: I believe XamDataChart is correct, but XamLegend likely not. As mentioned in another comment, we use the C# type names - the legend in our C# products is not "XamLegend" - it is just "Legend"

en/components/data-chart-navigation.md
gmurray81: this should not be XamDataChart by the time it reaches the angular output. I think something went wrong with the translation

gmurray81: in xplat-docfx it should be XamDataChart, but by the time it gets here it should not.

en/components/data-chart-legends.md
gmurray81: yes, that should be Legend

en/components/data-chart-navigation.md
gmurray81: oh nm, we are in xplat docfx atm

gmurray81: no, nm, I was right initially
we are not :)

gmurray81: It's seeming very much like the translation broke and stopped mapping the type names. This shouldn't be merged like this and we need to investigate.

gmurray81: This is ok, actually, it is inside a commented out area, so the translation is ignoring the code terms, accordingly.

PR #1694, opened 2019-12-04
All Data summary topic 

ddincheva:


en/components/grids_templates/summaries.md
```diff
@@ -279,6 +279,56 @@ export class HGridSummarySampleComponent implements OnInit {
 }
 ```
 }
+
+#### Custom summaries, which access all @@igComponent data
+ Now you can access all grid data inside the custom column summary. Two additional optional parameters are introduced in the IgxSummaryOperand `operate` method.
+As you can see in the code snippet below the operate method has the following tree parameters:
```
hanastasov: three

en/components/grids_templates/summaries.md
```diff
@@ -279,6 +279,56 @@ export class HGridSummarySampleComponent implements OnInit {
...
+- columnData - gives you an array that contains the values only for the certain column
```
hanastasov: I don`t think 'certain' suits here, I would change it to "current" or "given"

en/components/grids_templates/summaries.md
```diff
@@ -279,6 +279,56 @@ export class HGridSummarySampleComponent implements OnInit {
...
+- gridDataSource - gives you the whole grid data source
+- fieldName - current column field
+
+```typescript
+class MySummary extends IgxNumberSummaryOperand {
+    constructor() {
+        super();
+    }
+    operate(columnData: any[], allGridData = [], fieldName?): IgxSummaryResult[] {
```
hanastasov: the parameter is "allGridData", but allData is used afterwards

PR #1732, opened 2020-01-09
Changes from AngularDocFX_JP

ESShared:


zdrawku:
Yoshimi, this PR is part of the fix based on the Redirect issues

PR #1740, opened 2020-01-20
[DO NOT MERGE]Replaced kr domains to use infragistics.co.kr

### LINKED ISSUE #1742: Update docfx to version 2.4.6
Related to #1742

jasonzajac:
Updated template to replace old KR domains with the new one -- www.infragistics.co.kr

Closes #1742


HristoP96:
The version of the template which should be applied is 2.4.7

PR #1741, opened 2020-01-21
Update docfx-template to 2.4.6

HristoP96:


HristoP96:
This PR should be merged without updating the docfx-template, since the changes in version 2.4.6. are now on staging since current vNext version of the template is 2.4.7

PR #1789, opened 2020-02-24
Css variables default #6803

didimmova:
https://github.com/IgniteUI/igniteui-angular/issues/6803


en/components/themes/component-themes.md
```diff
@@ -180,25 +179,12 @@ The below sample uses the sample from the [View Encapsulation](#view-encapsulati
...
-}
-```
-As a bonus, any Ignite UI for Angular theme built with the `$igx-legacy-support` set to `false` will allow styling of components without the need to use Sass in your project. For instance the above could be achieved by setting the value of `--igx-avatar-initials-background` CSS variable to the desired color:
-
-```css
-/* app-avatar.component.css */
-
-:host ::ng-deep {
-    --igx-avatar-initials-background: royalblue;
+    @include igx-avatar($avatar-theme);
```
simeonoff: Please make sure this example shows how you can style view encapsulated components using CSS variable themes and without the use  of `::ng-deep`. In other words, rework the entire sample.

PR #1793, opened 2020-02-25
add range date picker section vNext

### LINKED ISSUE #1766: add section for range-date-picker in the date-picker topic
add section for range-date-picker with calendar and input in the date-picker topic
#1646

MonikaKirkova:
closes #1766


Lipata:
The comments in the sample should be addressed.

PR #1804, opened 2020-02-27
docs(*): update tags - add/remove new #1785

### LINKED ISSUE #1785: Reflect changes from 9.0.0 by adding/removing new tags
Remove new tags for components that are not new or updated with igniteui-angular@9.0.0 and add new for those that have updates that are entirely new.

Lipata:
Closes #1785

### DIFF
--- a/en/components/toc.yml
+++ b/en/components/toc.yml
@@ -8,7 +8,7 @@
   new: false
 - name: Angular Schematics & Ignite UI CLI
   href: general/cli-overview.md
-  new: false
+  new: true
   items:
     - name: Step-by-Step Guide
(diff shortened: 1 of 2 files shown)


en/components/toc.yml
```diff
@@ -876,13 +882,13 @@
...
       href: drop_down_virtual.md
-      new: true
+      new: false
 - name: Select
   href: select.md
   new: true
 - name: Autocomplete
   href: autocomplete.md
-  new: true
+  new: false
```
radomirchev: "Select" should be "true" - https://www.infragistics.com//angularsite/components/select.html#select-header--footer
We have new "header" and "Footer"

PR #1818, opened 2020-03-06
Add rtl support section

### LINKED ISSUE #1786: Add a section mentioning the RTL capabilities
We should mention everything related to the RTL configuration for our components.

For example:
- `dir="rtl"` should be set on the body or html element 

```
<body class="igx-typography default-theme" dir="rtl">
  <app-root></app-root>
</body>
```

- `$direction: rtl` should be set to the theme that is used

`@include igx-core($direction: rtl);`

My suggestion is to convert **framework-and-features** to .md file and use it for ARIA support and RTL configuration

Aleksandyr:
Closes #1786


zdrawku:
I've created a new PR by mistake https://github.com/IgniteUI/igniteui-docfx/pull/1821

PR #1826, opened 2020-03-09
docs(layout manager): update docs for layout manager

didimmova:
https://github.com/IgniteUI/igniteui-angular-samples/pull/1824


randriova:
#1853

PR #1932, opened 2020-05-12
Making editorial review on the Action Strip topic

StefanIvanov:


StefanIvanov:
@dkamburov as the author do you have any comments or we can go ahead with merging the PR?

PR #1939, opened 2020-05-14
Adds igxForOf dynamic height changing known limitation - 9.1

3phase:
Related issue: IgniteUI/igniteui-angular#7300 

### Checklist:
 
 - [ ] check topic's TOC/menu and paragraph headings
 - [ ] link to other topics using `../relative/path.md`
 - [ ] at the References section at the end of the topic add links to topics, samples, etc
 - [ ] reference API documentation instead of adding a section with API

<br />

 - [ ] use valid component names - [Data] Grid, `IgxSelectComponent`, `<igx-combo>`
 - [ ] use spell checker tool (VS Code, Grammarly, Microsoft Editor)
 - [ ] add inline `code blocks` for the names of classes / tags / properties
 - [ ] add language descriptor for the ```code blocks```
 - [ ] check broken links (use browser add-on)
 - [ ] check if sample is working and fully visible in the topic
 - [ ] check if sample is working and fully visible in the StackBlitz
 - [ ] check if code blocks match the code in StackBlitz demo

<br />

 - [ ] follow SEO guidelines to fill topic's metadata ([SEO guidelines](https://infragisticsinc297.sharepoint.com/:w:/g/Groups/marketing/EWz9InT4FDlErHCumxsKGY4Bd8H03yhRWxDFk47luRz-_Q?e=S5wWcx))

<br />

 - [ ] do not resolve requested changes (leave that to the reviewer)
 - [ ] add `pending-localization` label when the review of the PR is done
 - [ ] add a member from the localization team to translate it


en/components/for_of.md
```diff
@@ -212,6 +212,7 @@ In the example above, an `even` class is assigned to every even div element.
...
+| Content size changes at runtime are not anticipated by the `igxForOf` | When an element inside the `igxForOf` resizes at runtime, the recalculated width/height positions of the subsequent elements would not be in compliance with their actual ones. That behavior occurs because the directive renders and adjusts the scroll positions at once. In case any content size changes occur after that, they will not be taken into account.<br>A possible workaround of that problem is to set the dimensions of the projected container explicitly, so that any further resizes won't affect neither the display of the content, nor the calculations, performed by the `igxForOf`.
```
MayaKirova, line 215: Issue should be specific about "scrollTo" method as everything else works as expected:

`scrollTo` method does not work correctly when the content size of the rendered templates changes post initialization . |  When the elements inside the template have size that changes runtime after initialization (for example as a result of content projection, remote request resolution etc.) then the `scrollTo` method will not be able to scroll to the correct index. The method will scroll to the position of the index before the runtime size change occurs, hence the location will not be correct after the size is changed later. A possible workaround is to use templates that do not change their size based on their content if the content is loaded later.

skrustev:
Changes to the KR version shouldn't be made since they are handled by the localization team. Same for 9.0

skrustev:
Changes to the KR version shouldn't be made since they are handled by the localization team. Same for 9.0

3phase:
> Changes to the KR version shouldn't be made since they are handled by the localization team. Same for 9.0

@skrustev, thanks for your remark, it's resolved now.

PR #1977, opened 2020-06-01
Update licensing last section

### LINKED ISSUE #1946: Add new Licensing topic
- [x] Initial topic with step-by-step guide on how to migrate from trial to licensed package
- [x] Describe CLI automated migrations
- [x] Describe a way to integrate a Public CI process in a repository, which uses a licensed package (with access token) or a way to acquire a access token through infragistics.com user account, which later will be used to access the licensed package.

zdrawku:
Closes #1946 
@kdinev I didn't add a Github actions section as I am not familiar with this type of configuration.
@damyanpetev we've decided not to use the `echo` call for all configurations 

Note: This PR also addresses the issue that @radomirchev pointed out on the SoS - the link to data analysis sample was changed


en/components/general/ignite-ui-licensing.md
```diff
@@ -119,6 +119,77 @@ If you are upgrading from trial to licensed package and you are not using the au
...
+```cmd
+npm config set @infragistics:registry $(npmRegistry)
+```
+
+```cmd
+npm config set always-auth true --scope:@infragistics
+```
+
+```cmd
+npm config set _auth=$(token) --scope:@infragistics
```
damyanpetev, line 159: I'd advice against using those commands as a whole right now, since the scope parameter here does nothing (i.e. the set command doesn't accept such argument) and thus this ends up as a global setting which may or may not interfere with potential other credentials for npm registry/Azure artifacts/nexus etc. Can't be a 100% sure what, but https://stackoverflow.com/questions/49404870/how-to-set-auth-for-a-scoped-registry-in-npmrc seems to suggest those don't always work as expected and they don't get prefixed with the registry URI in the config

zdrawku: Wouldn't it be better to mention this with a Warning Note and leave it as is?

kdinev:
@zdrawku https://help.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions#example-using-a-private-registry-and-creating-the-npmrc-file - for github actions

en/components/general/ignite-ui-licensing.md
zdrawku: I've logged a separate [issue](https://github.com/IgniteUI/igniteui-angular-samples/issues/2060) for your comment. We are going to address it after further research on the subject.

PR #2236, opened 2020-09-17
[JP] Small changes in Breaking Changes topic.

randriova:


zdrawku (approved):
Merging this as the change already exists on internal xplat-docfx.

PR #3291, opened 2022-08-25
docs(theming): update documentation after theming package integration

didimmova:


en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
 
 For example: if you are updating from version 6.2.4 to 7.1.0 you'd start from the "From 6.x .." section apply those changes and work your way up:
 
+## From 14.1.x to 14.2.x
+### General
+- **Breaking Change** - All **igx** prefixes are replaced with **ig** prefixes. For example, **igx-**typography is now **ig-**typography.
```
simeonoff, line 56: We should clarify that this change only applies to global CSS variables for palette, elevations, typography, etc. As well as the igx-typography class. All component related variables retain the igx prefix.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+
+- Add this line in your `angular.json` file:
+```
+"stylePreprocessorOptions": {
+    "includePaths": ["node_modules"]
+}
+```
```
simeonoff, line 63: Where exactly in my angular.json file should I add this? Further clarification is needed.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+to access the `igniteui-theming package`, without writing the full path. You should have the igniteui-theming package installed in your `node_modules` folder first. If it's not installed automatically, use the `npm install igniteui-theming` command in your project directory.
+
+### Themes
+- **Breaking Change** - All `$grays` variable names are switched with `$gray`. <br />
```
simeonoff, line 67: This should be more specific. The name has changed in several places, not just the argument `$grays`. So mention that the $grays input argument must be changed in the `palette`, `color`, and `contrast-color` functions, but also that `--igx-grays-*` now is `--ig-gray-*`. All this will be much better explained using code samples to support the text description.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+If you're using the **color function** to get a gray shade, use it like this: `color($green-palette, 'gray', 800);`
+
+- **Breaking Change** - **Generating css variables** for the palette is now done by the **palette mixin**, not by the **palette-vars mixin**.
```
simeonoff, line 70:
```suggestion
- **Breaking Change** - **Generating CSS variables** for the palette is now done by the **palette mixin**, not by the **palette-vars mixin**.
```

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+
+- **Breaking Change** - The **palette function** now **requires the surface color** and the **grays parameter** is renamed to **gray**. The gray color can be generated automatically based on the surface color, or you can set one yourself. When you're generating a palette, you have to keep in mind that there are no longer default values for **info, success, error and warn** colors. You have to set the explicitly if you want to use them.
```
simeonoff, line 72:
```suggestion
- **Breaking Change** - The **palette function** now **requires the surface color** and the **grays parameter** is renamed to **gray**. The gray color will be generated automatically based on the surface color, or you can set it yourself. When you're generating a palette, you have to keep in mind that there are no longer default values for **info, success, error and warn** colors. They have to be set explicitly if you want to use them in custom created palettes.
```

This should also showcase how users can pick the info, success, error, and warn colors from an existing palette when creating a custom palette if they don't want to come up with the colors themselves.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+
+- **Breaking Change** - **The palette parameter is now removed from all component themes.** You can **use the palette mixin** to include your custom palette in your component. When making a custom theme for a component it's good to know that using the css-vars mixin is generating css variables only for those properties you've changed. Including the component mixin generates css variables for all properties in the component's theme.
```
simeonoff, line 74:
```suggestion
- **Breaking Change** - **The palette parameter is now removed from all component themes.** You can **use the palette mixin** to include your custom palette in your component. When making a custom theme for a component it's good to know that using the css-vars mixin is generating css variables only for the properties you've changed. The component mixin generates CSS variables for all properties in the component's theme and includes all styles for the component again, doubling the size of your CSS. The component-specific mixins should **only** be used if you've excluded the component theme from the global theme of styles (link to full article)[https://add-link-here].
```

simeonoff: Add a supporting code sample

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+
+### Typography
+- **Breaking Change** - The **type-style** mixin now doesn't accept type-scale as a parameter, only the categories names.
```
simeonoff, line 77:
```suggestion
- **Breaking Change** - The **type-style** mixin now doesn't accept type-scale as a parameter, only the category name.
```

Again, add a supporting code sample.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,35 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+
+### Elevations
+- **Breaking Change** - The **elevation function** now accepts only one argument - **$name - the elevation level**.
+- **Breaking Change** - The **elevations function** has been removed, you can now configure the elevation colors, using the `configure-elevations mixin.
+
```
simeonoff, line 82: Add supporting code samples.

simeonoff (changes requested):
Add supporting code blocks to the migration guide.

en/components/general/update-guide.md
```diff
@@ -51,6 +51,168 @@ Unfortunately not all changes can be automatically updated. Changes below are sp
...
+    ```sh
+    npm install igniteui-theming
+    ```
+
+    ```json
+        "build": {
+          "options": {
+            "stylePreprocessorOptions": {
+                "includePaths": ["node_modules"]
```
pmoleri, line 66: couldn't this be done in a migration?

simeonoff: We don't have migration instructions to do that.

pmoleri: Well, I remember that in our update to Angular@12 our `angular.json` was updated by Angular.
So, I figured the same should be possible for us, unless there's some kind of protection for it.
In that case it's more a question of if it's worth the effort or not.

simeonoff: Can't tell you if it's worth or not, but I don't think we have time to implement it before the release.

simeonoff (approved):

PR #3378, opened 2022-11-07
docs(qb): modifying SEO for QB topic

ChronosSF:


ChronosSF:
@aatanasov2 , this is what I came up with though it's not that much. Can't find Katya's GitHub handle to assign her to check it out.

aatanasov2 (approved):

PR #5891, opened 2024-07-17
Temp fix for ubuntu-only bug: localhost in code view URLs

turbobobbytraykov:


dobromirts (approved):

turbobobbytraykov:
No longer relevant. Fix in the gulp script took care of the problem so we are good on Ubuntu now.

